PR #8, opened 2017-04-12
Actually ignore warnings

kamalmarhubi:
Use the correct attribute, move it to the crate root. Also fix a couple
of legitimate warnings in cjson.


kamalmarhubi:
Actually I can do a better job here.

kamalmarhubi:
Or this is probably fine actually.

heartsucker:
Awesome. Thanks!

Issue #2, opened 2017-04-11
RSA key verification


heartsucker:
I know `ring` has some goodies in it, and at minimum we want to support RSASSA-PSS.

tonychain:
I would definitely suggest using *ring*. I am less enthusiastic about RSA (even with PSS) although if the goal of this project is to be a reusable TUF implementation I suppose it makes sense, and *ring* has a decent RSA implementation.

briansmith:
Note that the *ring* API is designed so that you can use RSA simply by swapping `signature::ED25519` in your code for `signature::RSA_PSS_2048_8192_SHA256`. So probably the best thing to do is to parameterize all the signature checking logic within rust-tuf with a `&'static signature::VerificationAlgorithm` parameter.

heartsucker:
Note: This (very much most likely) depends on #3 or at least a better implementation of the `cjson` bits I implemented.

heartsucker:
Update: progress can be found in the `rsa` branch.

heartsucker:
Partially blocked by https://github.com/briansmith/ring/issues/530

heartsucker:
Adding to version 0.1.0 because this is needed for testing something else.

PR #27, opened 2017-04-17
Move from rust-crypto to ring

vks:
This replaces all use of rust-crypto with ring. I think ring is a better choice, because it is better documented, more actively developed and more minimal. Besides, this is a prerequisite for #2.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,5 +1,5 @@
 extern crate chrono;
-extern crate crypto;
+extern crate ring;
 extern crate itoa;
 #[macro_use]
 extern crate log;
@@ -10,6 +10,7 @@ extern crate serde_derive;
 #[macro_use]
 extern crate serde_json as json;
 extern crate url;
+extern crate untrusted;
 
 mod cjson;
 mod metadata;
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -1,7 +1,8 @@
 use chrono::{DateTime, UTC};
-use crypto::digest::Digest;
-use crypto::ed25519;
-use crypto::sha2::Sha256;
+use ring;
+use ring::digest::{digest, SHA256};
+use ring::signature::{verify, ED25519};
+use untrusted::Input;
 use json;
 use rustc_serialize::hex::{FromHex, ToHex};
 use serde::de::{Deserialize, Deserializer, Error as DeserializeError};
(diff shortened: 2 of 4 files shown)


src/metadata.rs
```diff
@@ -565,7 +562,10 @@ impl SignatureScheme {
...
         match self {
             &SignatureScheme::Ed25519 => {
-                if ed25519::verify(msg, &pub_key.0, &sig.0) {
+                if ring::signature::verify(
+                    &ED25519,
+                    Input::from(&pub_key.0), Input::from(msg), Input::from(&sig.0)
+                ).is_ok() {
                     Ok(())
                 } else {
                     Err(Error::VerificationFailure("Bad signature".into()))
```
briansmith: Usually there is a better way of doing things than using `is_err()`. For example:

`ring::signature::verify(&ED25519, Input::from(&pub_key.0), Input::from(msg), Input::from(&sig.0), Input::from(msg), Input::from(&sig.0)).map_err(|_| Error::VerificationFailure("Bad signature".into()))`.

src/tuf.rs
```diff
@@ -292,32 +291,32 @@ impl Tuf {
...
         }
     }
 
-    fn read_and_verify<R: Read, D: Digest>(input: &mut R,
-                                           digest: &mut D,
-                                           size: i64,
-                                           expected_hash: &[u8])
-                                           -> Result<(), Error> {
+    fn read_and_verify<R: Read>(input: &mut R,
+                                mut context: Context,
```
briansmith: IMO it would be clearer to qualify it as `digest::Context` because there are (will be) several `Context` types.

src/lib.rs
```diff
@@ -1,5 +1,5 @@
 extern crate chrono;
-extern crate crypto;
+extern crate ring;
```
heartsucker: Pedantic: can you move this down to keep the list of crates in alphabetical order?

src/metadata.rs
```diff
@@ -1,7 +1,8 @@
 use chrono::{DateTime, UTC};
-use crypto::digest::Digest;
-use crypto::ed25519;
-use crypto::sha2::Sha256;
+use ring;
```
heartsucker: Pedantic: Can you move these to keep the imports in alphabetical order (just the crate, not the modules/structs).

vks:
I think I addressed all the comments.

heartsucker:
Awesome. Thanks @vks!

PR #40, opened 2017-05-16
Upgrade to serde 1.0

### LINKED ISSUE #34: Upgrade to serde 1.0
And remove the hard dependency on `chrono = 0.3.0`.

vks:
The trait `Metadata` now implies `DeserializeOwned` instead of `Deserialize`.

Fixes #34.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ required-features = [ "cli" ]
 cli = [ "clap" ]
 
 [dependencies]
-chrono = { version = "= 0.3.0", features = [ "serde" ] }
+chrono = { version = "0.3", features = [ "serde" ] }
 clap = { version = "2.23", optional = true }
 data-encoding = "2.0.0-rc.1"
 env_logger = "0.4.2"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -29,17 +29,17 @@ required-features = [ "cli" ]
 cli = [ "clap" ]
 
 [dependencies]
-chrono = { version = "= 0.3.0", features = [ "serde" ] }
+chrono = { version = "0.3", features = [ "serde" ] }
```
heartsucker: Can you change this to `0.3.1` since `0.3.0` isn't compatible?

heartsucker:
Just FYI, I still can't get this to pass the tests on Windows, so ignore the appveyor CI failures when pushing (unless you want to fix it in another PR).

vks:
I corrected the chrono version.

Not sure about the the windows failures. It seems to be some filename issue: `"The filename, directory name, or volume label syntax is incorrect."`

heartsucker:
@vks It has someting to do with the URL to win-path conversion (I think), but like I said. you can ignore that. Anyway. Thanks for the contribution.

PR #47, opened 2017-05-29
Link to what TUF is in the README

sgrif:
We shouldn't assume that everyone who will stumble on this crate is familiar with what TUF is, and we can save them a frustrated google search.


heartsucker:
Good point. Thanks!

Issue #99, opened 2017-07-05
tuf/remote: HTTP client with custom/authentication headers


lucab:
I was looking into the feasibility of integrating this crate into a docker client, as the docker registry has a TUF-based trusted distribution extension called notary.

One initial issue I'm seeing is that the docker notary endpoints require authentication via HTTP header, but I'm not seeing any direct way of setting custom headers on the internally-built request. Do you have any suggestions in that direction? Should `ConfigBuilder` grow some way to take a third-party function which manipulate the `RequestBuilder` before sending it?

heartsucker:
This likely wouldn't be compatible with Notary since the spec has moved on from a rigid definition of metadata fields, encodings, and data interchange formats. One consumer of this lib already had problems because this deviates from the Python reference implementation: https://github.com/heartsucker/rust-tuf/issues/93

It is possible that Notary and the Python impl both followed and old version of the draft to the T and are compatible, but that likely would not be the case between future implementations.

However, if in general you want to be able to set custom headers, that should be possible.

lucab:
Thanks for the forewarning, I literally just started scratching the surface of this so it may as well turn out to be non-compatible. I'll report back once I'll get interesting datapoints.

heartsucker:
It actually looks like `rustup` is pinned to `hyper 0.9.8` right now, so it would probably make more sense to have `rust-tuf` and `rustup` both meet a `0.10` instead of advancing them both to `0.11`.

I opened an upstream issue to address that: https://github.com/rust-lang-nursery/rustup.rs/issues/1195

Issue #93, opened 2017-07-03
Minimal working example


nbigaouette-eai:
I'm looking into tuf/rust-tuf for some update procedures.

For now I'm failing to create a minimal working example and I'm not sure if it's just me or if there is a bug in the documentation that needs fixing. rust-tuf's src/lib.rs contains an example with a function `get_original_root()` that simply is `unimplemented!()`. I'm trying to fill the gap here.

First, I have followed the TUF documentation to generate a directory structure to be served according to the documentation in https://github.com/theupdateframework/tuf/blob/develop/tuf/README.md. In a nutshell, the steps are:
1. Generate RSA keys (root, targets, snapshot, timestamp);
1. Create repository metadata;
1. Create the actual repository.

Step 1. is trivial, simply calling `tuf.repository_tool.generate_and_write_rsa_keypair()` for each key. Step 2. uses `create_new_repository()` and adds the keys to the repo. Step 3. walks the repository directory and signs files using `add_targets()`.

The problem I am facing comes from the call to rust-tuf's [`tuf::interchange::JsonDataInterchange::from_reader()`](https://github.com/heartsucker/rust-tuf/blob/c3321bf1295be66df46dcfb0440df491f2cf583f/src/lib.rs#L37-L38). Since the `get_original_root()` function in `unimplemented!()`, I had to adapt it. It's returning a `File`, so I let it return `File::open("repository/metadata/root.json").unwrap()` where the json file was created by step 3 above.  The `from_reader()` call fails though with a deserialization error:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Encoding("JSON: ErrorImpl { code: Message(\"missing field `key_id`\"), line: 7, column: 3 }")', src/libcore/result.rs:859
```
I checked the files generated by `tuf.repository_tool`, and none contains the string `key_id`. The rust-tuf [struct `Signature`](https://github.com/heartsucker/rust-tuf/blob/c3321bf1295be66df46dcfb0440df491f2cf583f/src/crypto.rs#L466) seems to have one such field and is `Deserialize` so maybe this is where the error comes from?

[...]

PR #108, opened 2017-07-09
repository/http: add a custom pre-send hook

lucab:
This adds an optional user-provided boxed closure to `HttpRepository`
which can be triggered to customize an HTTP request before sending it.

Closes https://github.com/heartsucker/rust-tuf/issues/99

### DIFF
--- a/src/repository.rs
+++ b/src/repository.rs
@@ -1,6 +1,7 @@
 //! Interfaces for interacting with different types of TUF repositories.
 
 use hyper::{Url, Client};
+use hyper::client::RequestBuilder;
 use hyper::client::response::Response;
 use hyper::header::{Headers, UserAgent};
 use hyper::status::StatusCode;
@@ -289,6 +290,8 @@ where
     }
 }
 
+/// A custom hook to be invoked before sending HTTP requests.
+pub type HttpHook = Box<for<'a> Fn(RequestBuilder<'a>) -> Result<RequestBuilder<'a>>>;
 
 /// A repository accessible over HTTP.
 pub struct HttpRepository<D>
@@ -297,6 +300,7 @@ where
 {
     url: Url,
     client: Client,
+    http_hook: Option<HttpHook>,
     user_agent: String,
     _interchange: PhantomData<D>,
 }
@@ -317,9 +321,40 @@ where
         HttpRepository {
             url: url,
             client: client,
+            http_hook: None,
             user_agent: user_agent,
             _interchange: PhantomData,
         }
+    }
+
+    /// Set an hook function to be invoked before sending an HTTP request.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # extern crate hyper;
+    /// # extern crate tuf;
+    /// # fn main() {
+    /// use hyper::header::{Authorization, Bearer};
+    /// use tuf::repository::{HttpHook, HttpRepository};
+    /// use tuf::interchange::JsonDataInterchange;
+    ///
+    /// let cl = hyper::Client::new();
+    /// let token = "<opaque blob>".to_string();
+    /// let url = "https://example.com".parse().unwrap();
+    ///
+    ///
+    /// let mut repo = HttpRepository::<JsonDataInterchange>::new(url, cl, None);
+    /// let hook: HttpHook = Box::new(move |r| {
+    ///     let auth = Authorization(Bearer { token: token.clone() });
+    ///     Ok(r.header(auth))
+    /// });
+    /// repo.http_hook(Some(hook));
+    /// # }
+    /// ```
+    pub fn http_hook(&mut self, hook: Option<HttpHook>) -> &Self {
+        self.http_hook = hook;
+        self
     }
 
     fn get(&self, components: &[String]) -> Result<Response> {
@@ -333,7 +368,10 @@ where
             })?
             .extend(components);
 
-        let req = self.client.get(url.clone()).headers(headers);
+        let mut req = self.client.get(url.clone()).headers(headers);
+        if let Some(ref hook) = self.http_hook {
+            req = hook(req)?;
+        }
         let resp = req.send()?;
 
         if !resp.status.is_success() {
@@ -525,9 +563,11 @@ where
 
 #[cfg(test)]
 mod test {
+    use std::panic;
     use super::*;
     use tempdir::TempDir;
     use interchange::JsonDataInterchange;
+    use hyper::header::Authorization;
 
     #[test]
     fn ephemeral_repo_targets() {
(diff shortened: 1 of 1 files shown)


src/repository.rs
```diff
@@ -317,11 +321,18 @@ where
...
             url: url,
             client: client,
+            http_hook: None,
             user_agent: user_agent,
             _interchange: PhantomData,
         }
     }
 
+    /// Set an hook function to be invoked before sending an HTTP request.
+    pub fn set_http_hook(&mut self, hook: Option<HttpHook>) -> &Self {
```
heartsucker: Pedantic: I'm trying to avoid the Java `set_xxx` convention, so could you change this to just `http_hook`.

Less pedantic: Could you put a short example in the docs showing how someone might use this to set auth headers (or whatever your use case was). It possible the description isn't enough.

heartsucker (changes requested):

src/repository.rs
```diff
@@ -591,4 +606,23 @@ mod test {
         read.read_to_end(&mut buf).expect("read target");
         assert_eq!(buf.as_slice(), data);
     }
+
+    #[test]
+    #[should_panic]
```
heartsucker: I also wonder if this could use a `panic::catch_unwind` just on the final step to ensure we only catch the correct panic.

heartsucker (changes requested):

src/repository.rs
lucab: * I also don't like getters&setters either, but I thought the non-prefixed one is typically used for getters not for setters. But I will amend it.
 * Sure. Is something similar to the test-case (minus the panic part) ok?

src/repository.rs
lucab: Uhm, I never tried this but it seems feasable and reasonable, I'll have a look at it.

lucab:
Full disclosure: I didn't test it in my docker usecase yet as I coded while travelling/offline and you forewarned me it probably won't interoperate anyway, but I still wanted to code this and get a review. Do you prefer to wait for a full roundtrip test in my scenario before landing this?

src/repository.rs
```diff
@@ -289,6 +290,8 @@ where
     }
 }
 
+/// A custom hook to be invoked before sending HTTP requests.
+pub type HttpHook = Box<for<'a> Fn(RequestBuilder<'a>) -> RequestBuilder<'a>>;
```
heartsucker: I was thinking about this some more. I think the signatures should be

```rust
// non-public type
type HttpHook = Box<for<'a> Fn(&mut RequestBuilder<'a>) -> Result<RequestBuilder<'a>, Into<Error>>; 
```

And the function docs needs to say that it **must not** panic. I think this is better because I would rather consumers of this lib not introduce panicking code. On top of this, I think it would be good to wrap the call site itself in a `panic::catch_unwind` and returning `Error::Opaque("Hook panicked.".into())` to contain the panic and stop it from spreading elsewhere in the code.

heartsucker (changes requested):

heartsucker:
> Do you prefer to wait for a full roundtrip test in my scenario before landing this?

Yes. It would be good to ensure that it suits your use case before we put this into the lib. Transcribing your use case into the docs would be helpful so that other people see how to use the hook.

lucab:
Ack, I will then proceed to address the comment from the review, push it, make sure it can authn to a docker notary and then report back. The full scenario is a bit too large to fit (multi-steps flow to retrieve an authn token), but it can be simplified in a bearer token header (similar to the testcase).

heartsucker:
Ok, so maybe not the *whole* case then, but a partial case would be nice. I'm also not 100% sure about what do with my `panic::catch_unwind` proposal as I generally just completely avoid panics in my code, so I'm not sure if what I said is even idiomatic or makes sense. Though I think it does. It would be nice if someone like @brson or a more seasoned Rust dev could give some insight here.

Edit: see other comments. no `catch_unwind`s, plz.

src/repository.rs
heartsucker: Oops. Just saw this. I'm not using getters since this is a builder and presumably people know what they're putting in before they do, so I won't need to support both.

And yeah, something similar to your test case but without the panic and with an actual manipulation.

src/repository.rs
heartsucker: Ok, so I did my research and realized I misunderstood `catch_unwind`. Don't wrap call site, actually. Turns out that's a bad idea.

heartsucker:
Heya. Thought about this some more. I have the feature in here were I set the lib version in the user agent, and your method would allow someone to overwrite that. This is actually pretty important telemetry for ensuring clients in the field aren't cut off from updates because of this lib making breaking changes.

Also, it should remove the possibility to set the body (since we're only `GET`-ing). I almost think it would be better to give a `Result<Headers, Into<Error>>`. Though someone might need to set query string params as well, so we might need to get back a `Result<(HashMap<String, String>, Headers), Into<Error>>` and now we're getting very complicated.

I'm more hesitant about this PR now that I've run through it a few times.

lucab:
Yes, this is a pre-send hook which means that you (as a library author) trust me (as the consumer/service owner) to know better than you about what my environment quirks are, and override some of your decisions. Coming back to your examples, my TUF server may have ~~weird~~over-smart LB in front which requires exact hardcoded user-agent, or some body content in the GET. As unfortunate, extreme and wrong as it sounds, this unfortunately may happen when interacting with somebody else's infra.

Back to a technical point, I think that two possible approaches are:
 * embrace the fact that users may need to override your choices, and let them do it (current PR)
 * try to undo some of the user changes, by invoking the hook earlier and later undoing what you don't like

I find the latter a bit wrong as a library consumer, but if you are mostly concerned with an untouched user-agent I think I can swap the hook order and the clean+set the user-agent. I'd still prefer current approach, though.

> Also, it should remove the possibility to set the body (since we're only GET-ing)

Quoting from memory without RFC at hand, this is not an invariant. GETs *may* have body content and servers *should* ignore it.

> Though someone might need to set query string params as well

This is out of scope from this PR. If somebody needs this, we first have to capture the requirements (is the query dynamic/static? Is it homogeneous across all endpoints/actions?) and then think about how to expose it. I'm not trying to tackling it here, as I wouldn't be able to make the right call.

heartsucker:
Ok, those are fair points. I guess as lib author I like to limit the number possible dumb things someone could do. If this PR works for you, then if you address the couple of small issues I pointed out earlier, I'll merge it. I might hide it behind an unstable flag, however.

lucab:
Good news: this works, tested with the gist at https://gist.github.com/lucab/73a3519e2086a27cd21cbbf50abed6ff (ignore the fact it is a silly example, I just needed some debug-printf for the final manifest hash).
Bad news: as forewarned I hit https://github.com/heartsucker/rust-tuf/issues/93#issuecomment-312704968, more on this in a separate thread.

For review: second iteration addressing your previous comments, PTAL. Changes:
 * renamed method and tweaked to return a Result type
 * added a proper example based on the above gist with an opaque bearer token
 * wrapped the panic-ing part of the test in its own catch block

lucab:
Bump, what should we do with this? I amended as suggested, but it turns out in the end I'm not going to actively consume this or this library. Do you think it is worth anyway to land it for other authenticated TUF endpoints or should I just close it?

heartsucker:
Yeah, sorry I haven't been able to look at this. I think it's a good idea, and I think a future version of `rust-tuf` will include it, but I think we will switch to `hyper` `0.11.x` before this is actually needed. I think we can close it for now and figure out what else to do later. If you agree, I'll let you close.

lucab:
Ack, closing it. Yes, I hope reqwest at some point will grow direct support for this kind of hooks.

PR #110, opened 2017-07-11
cargo: bump ring to 0.11

lucab:
This updates to latest ring to interop with crates depending on newer
versions. In particular this avoid linking errors like:
```
error: native library `ring-asm` is being linked to by more than one
version of the same package, but it can only be linked once; try updating
or pinning your dependencies to ensure that this package only shows up once

  ring v0.9.7
  ring v0.11.0
```


heartsucker:
Danke!

PR #111, opened 2017-07-12
Add comments, tests, and broken re-impl of `matches_chain`

jamesmunns:
Add some test cases, and some comments. At least one of them needs correction.

In particular, I think either my understanding of the match algorithm is incorrect, or there may be some edge cases you are not handling that are confusing me.

Check out the test cases I added, and see if any of them look weird to you. If not, check the comments I added, and see if I am wrong somewhere.

I think I am close with my re implementation, but it is failing some test cases that the current impl passes. It still isn't great, no reallocs/clones, but it does have a worst case of `O(N*M + (N-1)*M*M)`ish (cubic), but tbh I think the worst case will always be cubic (unless I am missing a rule somewhere)


heartsucker:
> 4:11 AM

Lol dude. You really did take my challenge. :D

So yes, order matters. The idea being in TUF you can delegate responsibility. For example:

```
Top-Level-Targets:
  Targets:
    null
  Delegations:
    Role-A:
      - foo
      - bar/

Role-A:
  Targets:
    - foo
  Delegations:
    Role-B:
      - bar/

Role-B:
  Targets:
      null
  Delegations:
    Role-C:
      - bar/baz/
      - quux

Role-C:
  Targets:
    - bar/baz/wat
    - quux
```

In this case `C` is authorized for `bar/baz/wat` because its parent `B` is authorized for `foo/bar/` and *its* parent `A` is authorized for `bar/`. `quux` is not authorized by because the chain is "broken" because it wasn't authorized at the top level to `A`.

My "auth" scheme is either:

- exactly one file
- a dir, and everything under it recursively

And it can be a list containing many of both types.

And yeah, missing some edge cases in my tests, but this was enough for me to call "not borked to all hell" before I went to bed last night. I'll take a look at this today.

heartsucker:
I added some more test cases on the ride to work and it looks like neither or ours work. :(

heartsucker:
Yo @jamesmunns. I'm going to close this since I couldn't get yours to work with one of the cases ,and I have redone it on the `delegations` branch. Once I get all the tests to pass, we can take a crack at doing less cloning and heaping.

PR #130, opened 2017-09-01
split path in TargetPath and VirtualTargetPath

### LINKED ISSUE #129: Support for "virtual file sytems" to allow for target naming that is non-heirarchical
From this comment: https://github.com/rust-lang-nursery/rustup.rs/issues/1217#issuecomment-321271193

The idea being that targets might not support a naming scheme that matches what TUF expects. This would allow a user to provide functions that map "real" names to TUF friendly names and back.

heartsucker:
to allow for hierarchical addressing of targets when their real names
don't fit such a schema

e.g., foo-bar-1.2.3 could be addressed as foo/bar/1.2.3 so that foo
could delegate bar and so on

this happens because delegations are done on an exact target basis or
recursively or everything under a "dir" like "foo/"

fixes #129, assists with https://github.com/rust-lang-nursery/rustup.rs/issues/1217

@alexcrichton, since this is something you requested, can you take a peek and let me know if this looks like a good solution.

### DIFF
--- a/src/bin/tuf.rs
+++ b/src/bin/tuf.rs
@@ -1,6 +1,5 @@
 extern crate clap;
 extern crate ring;
-extern crate tuf;
 
 use clap::{App, AppSettings, SubCommand, Arg, ArgMatches};
 use ring::rand::SystemRandom;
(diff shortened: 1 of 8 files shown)


alexcrichton:
Nice!

Issue #131, opened 2017-09-01
Appveyor errors when persisting temp files


heartsucker:
Possibly related to permissions?

https://ci.appveyor.com/project/heartsucker/rust-tuf/build/302/job/rg32sb094w5oxydn

```
---- repository::test::file_system_repo_targets stdout ----
	thread 'repository::test::file_system_repo_targets' panicked at 'called `Result::unwrap()` on an `Err` value: Opaque("Error persisting temp file: PersistError { error: Error { repr: Os { code: 5, message: \"Access is denied.\" } }, file: NamedTempFile(\"C:\\\\Users\\\\appveyor\\\\AppData\\\\Local\\\\Temp\\\\1\\\\rust-tuf.V4hIuJmhxxk0\\\\temp\\\\.tmpoqHfon\") }")', src\libcore\result.rs:860:4
stack backtrace:
   0:   0xb6b03c - std::panicking::default_hook::{{closure}}::h48c07d1bc07c0089
   1:   0xb6ad55 - std::panicking::default_hook::h72dcd99c4287a2b2
   2:   0xb6b3dc - std::panicking::rust_panic_with_hook::h5b41b898e5797800
   3:   0xb6b2fe - std::panicking::begin_panic_new::hcf5fda792d4e8857
   4:   0xb6b242 - std::panicking::begin_panic_fmt::h673341051b2e87ff
   5:   0xb6b1c9 - rust_begin_unwind
   6:   0xb7df4a - core::panicking::panic_fmt::hfda8af447f9f88e7
   7:   0x4f924c - core::result::unwrap_failed::hd388f4b0cb245d3c
                       at C:\projects\rust\src\libcore/macros.rs:41
   8:   0x4c6580 - <core::result::Result<T, E>>::unwrap::hfe9570c1481489b0
                       at C:\projects\rust\src\libcore/result.rs:738
   9:   0x60eafd - tuf::repository::test::file_system_repo_targets::h026b0e12dfcc62e6
                       at src/repository.rs:557
  10:   0x9f540e - <F as test::FnBox<T>>::call_box::h05ad4323db8f2f84
  11:   0x9e862d - std::panicking::try::do_call::h894b4439d97b361e
  12:   0xb6e42a - __rust_maybe_catch_panic
  13:   0x9e7b00 - std::sys_common::backtrace::__rust_begin_short_backtrace::hf05802d444e79d0c
  14:   0x9e860b - std::panicking::try::do_call::h7013ce95188dd650
[...]

PR #133, opened 2018-07-12
bump dependencies

erickt:
This updates the dependencies the following dependencies to the latest version: derp, itoa, log, tempfile, lazy_static, maplit

The only remaining outdated dependencies are on hyper (which is going to be a doozy to update) and untrusted, which is blocked on ring 0.13's release.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,18 +23,18 @@ path = "./src/lib.rs"
 [dependencies]
 chrono = { version = "0.4", features = [ "serde" ] }
 data-encoding = "2.0.0-rc.2"
-derp = "0.0.4"
+derp = "0.0.10"
 hyper = "0.10"
-itoa = "0.3"
-log = "0.3"
+itoa = "0.4"
+log = "0.4"
 ring = { version = "0.12", features = [ "rsa_signing" ] }
 serde = "1"
 serde_derive = "1"
 serde_json = "1"
-tempfile = "2.1"
+tempfile = "3"
 untrusted = "0.5"
 
 [dev-dependencies]
-lazy_static = "0.2"
-maplit = "0.1"
+lazy_static = "1"
+maplit = "1"
 tempdir = "0.3"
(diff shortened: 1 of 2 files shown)


heartsucker:
Hey thanks for this. I'll probably eventually move to `reqwest` over `hyper` because it's easier to use an I don't really need to worry about async operations since use of TUF is inherently a bit blocking. Are you on a Windows machine, because debugging these appveyor errors is going to be a total pain. It looks like it's just temp file errors in one of the tests.

erickt:
Hello! Since I have your attention, thanks for the library. We are trying to use it for the [fuchsia](https://fuchsia.googlesource.com/docs/+/HEAD/README.md)’s update system, so you’ll be seeing a bit more of me :)

Speaking of which, request would be fine (and I might just do that switch) or staying on hyper could work too. Fuchsia is mostly all async, so there may be some benefits to us to switch over to a more futures based model. Especially with async/await coming soon (which is being implemented by a coworker of mine).

Anyway,  I did happen to have a windows machine to test this, but I didn’t try it on the gnu runtime. I can poke at this new error later on today.

heartsucker:
I think that's totally fair to want `hyper` and `futures` for this, and my suggestion for `reqwest` was more meant to make the changes easier. That might be a good first step just because it would be considerably more fiddling to get this async.

As for the error, I actually suspect it has something to do with appveyor and not that crate being incompatible with windows. I'll see if I can figure something out this weekend too.

Also since you're planning on implementing it, please stay in touch since my original plan of getting this in to `rustup` kinda got halted because I kept banging my head into walls while integrating it (more my fault as a dev than this lib's). I did integrate it into a demo CLI at some point just to prove it works, so if it needs tweaks or additions, I'm all ears.

heartsucker:
Apparently I already saw this error in #131. :roll_eyes:

tarcieri:
For what it's worth, `reqwest` has an obnoxious number of dependencies. I've removed it from a few security-critical projects just to cut down on that:

https://twitter.com/bascule/status/984627292470050817

heartsucker:
That is a super fair reason to not use `reqwest`. Scrap that idea :)

heartsucker (approved):

heartsucker:
I merged this even with the appveyor failure because that is clearly unrelated to this PR. Thanks for the contribution. :)

PR #135, opened 2018-07-17
Fix tempfile test on windows

erickt:
The tempfile library opens files on windows in a mode that prevents the file from being opened multiple times. This fixes the test by just making sure to close the file before we try to open it again.


heartsucker (approved):
Wonderful. Thank you!

PR #140, opened 2018-07-17
Convert Repository types to use interior mutability

erickt:
This changes `Repository` to use interior mutability in order to store state. At the moment, `EphemeralRepository` is the only repository type that stores state. By switching over to internally using `Arc<Mutex<...>>>` to wrap the state, it allows repositories to be used concurrently across threads. This trait will be especially important as we transition over to using asynchronous io with Hyper 0.11+.

Another approach could be to wrap the repositories in `Client` with `Arc<Mutex<...>>`, but I'm worried that might lead to deadlocks with asynchronous use of the repositories.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -133,7 +133,7 @@ where
     ///
     /// **WARNING**: This method offers weaker security guarantees than the related method
     /// `with_root_pinned`.
-    pub fn new(config: Config<T>, mut local: L, remote: R) -> Result<Self> {
+    pub fn new(config: Config<T>, local: L, remote: R) -> Result<Self> {
         let root = local
             .fetch_metadata(
                 &MetadataPath::from_role(&Role::Root),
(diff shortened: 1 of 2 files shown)


heartsucker (changes requested):
There's a few place where you `unwrap` on `PoisonError`. Is there a reason we're not trying to recover and `Err`? Are we assuming the data in the `Arc` is fried and just panicking everywhere until eventually this boils up to the main thread and the system exits?

If so, could you add a note to those unwraps since I have otherwise (or so I think) been rather careful to write non-panicking code.

erickt:
I'm not sure if we can really recover from [PoisonError](https://doc.rust-lang.org/std/sync/struct.Mutex.html#poisoning) in this case, since we're wrapping a plain old `HashMap`, which could get in a bad state if it was partially modified before a panic happened. If we switched over to a concurrent map, then it'd probably be robust, but I'm not familiar with any of the available libraries.

And sure, I can add some notes. Unfortunately though, in my hyper branch it looks like we're going to have to throw a bunch more things in `Arc`s, which means more possible sources of panics.

heartsucker:
Ok, that's what I thought. If there's going to a more panics because of concurrency related errors, then I think it's fine to leave this as is (uncommented unwraps).

heartsucker:
Boop for rebase.

heartsucker:
(whoops)

erickt:
And rebased.

heartsucker (approved):

heartsucker:
Danke!

PR #154, opened 2018-08-03
Reformat with rustfmt 0.8.2-stable (08da30d 2018-06-06)

erickt:
rustfmt-preview was bumped to 0.8.2, so this reformats the source with it.

### DIFF
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -15,12 +15,14 @@ use interchange::DataInterchange;
 use shims;
 use Result;
 
+#[cfg_attr(rustfmt, rustfmt_skip)]
 static PATH_ILLEGAL_COMPONENTS: &'static [&str] = &[
     ".", // current dir
     "..", // parent dir
          // TODO ? "0", // may translate to nul in windows
 ];
 
+#[cfg_attr(rustfmt, rustfmt_skip)]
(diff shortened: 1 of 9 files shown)


src/metadata.rs
```diff
@@ -24,77 +24,19 @@ static PATH_ILLEGAL_COMPONENTS: &'static [&str] = &[
...
-    "\u{017}",
-    "\u{018}",
-    "\u{019}",
-    "\u{01a}",
-    "\u{01b}",
-    "\u{01c}",
-    "\u{01d}",
-    "\u{01e}",
-    "\u{01f}",
+    "<", ">", "\"", "|", "?", "*", // control characters, all illegal in FAT
```
heartsucker: Can you reformat these by hand to put the comments over the items they point to. The FAT one needs to drop down a line, and the others should reformatted for consistency.

heartsucker (changes requested):

src/metadata.rs
erickt: done, turns out it's easy to get rustfmt to ignore this with `#[cfg_attr(rustfmt, rustfmt_skip)]`.

heartsucker (approved):

PR #155, opened 2018-08-03
Atomically write metadata to file system

erickt:
Before this patch, `FileSystemRepository::store_metadata` would
stream writes into the destination. If the system crashed,
it's possible the metadata could be truncated. This refactors
the code to atomically write metadata to avoid this happening.

Note that this changes the behavior of
`FileSystemRepository::store_target`. It used to store the
temporary files in a `$REPO/temp` directory, but instead
this moves the temp files into the same directory as the
destination file. This was done to avoid the case where
`temp/` was stored on a separate filesystem than the
destination. If that were the case,
`NamedTemporaryFile::persist` would not be atomic.

### DIFF
--- a/src/repository.rs
+++ b/src/repository.rs
@@ -5,10 +5,10 @@ use hyper::header::{Headers, UserAgent};
 use hyper::status::StatusCode;
 use hyper::{Client, Url};
 use std::collections::HashMap;
-use std::fs::{self, DirBuilder, File};
-use std::io::{Cursor, Read, Write};
+use std::fs::{DirBuilder, File};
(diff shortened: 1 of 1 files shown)


heartsucker (approved):

tarcieri:
FYI, there's a crate for this: https://crates.io/crates/atomicwrites /cc @untitaker

erickt:
Thansk @tarcieri! I'll check it out.

untitaker:
fwiw the tempfile crate seems to be more actively maintained. I added a link in my readme

PR #157, opened 2018-08-09
Clients should locally save metadata

### LINKED ISSUE #153: Client never saves remote metadata locally
It looks like `Client` never actually saves metadata locally. We should fix this so we don't have to hit the remote host all the time.

erickt:
This patch merges `Client::update_local` and `Client::update_remote` in order to update `Tuf` following the pattern as described in the [spec section 5.1](https://github.com/theupdateframework/specification/blob/master/tuf-spec.md#the-client-application).

One potential problem though is I'm not sure how we can atomically update the local repository and the `Tuf` object, so for ease of implementation, this patch will update `Tuf` and could potentially error out writing the metadata to the local repository. We could instead implement some sort of rollback mechanism to `Tuf`, but I figure we can do that later if it's necessary.

Closes #153.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -43,18 +43,18 @@
 //!         local,
 //!         remote,
 //!     ).unwrap();
-//!     let _ = client.update_local().unwrap();
-//!     let _ = client.update_remote().unwrap();
+//!     let _ = client.update().unwrap();
 //! }
 //! ```
 
 use std::io::{Read, Write};
 
+use chrono::offset::Utc;
 use crypto::{self, KeyId};
 use error::Error;
 use interchange::DataInterchange;
 use metadata::{
-    Metadata, MetadataPath, MetadataVersion, Role, RootMetadata, SnapshotMetadata,
+    Metadata, MetadataPath, MetadataVersion, Role, RootMetadata, SignedMetadata, SnapshotMetadata,
     TargetDescription, TargetPath, TargetsMetadata, VirtualTargetPath,
 };
 use repository::Repository;
@@ -166,22 +166,31 @@ where
         I: IntoIterator<Item = &'a KeyId>,
         T: PathTranslator,
     {
+        let root_path = MetadataPath::from_role(&Role::Root);
+
         let root = local
             .fetch_metadata(
-                &MetadataPath::from_role(&Role::Root),
+                &root_path,
                 &MetadataVersion::Number(1),
                 &config.max_root_size,
                 config.min_bytes_per_second,
                 None,
             )
-            .or_else(|_| {
-                remote.fetch_metadata(
-                    &MetadataPath::from_role(&Role::Root),
+            .or_else(|_| -> Result<SignedMetadata<_, RootMetadata>> {
+                // FIXME: should we be fetching the latest version instead of version 1?
+                let root = remote.fetch_metadata(
+                    &root_path,
                     &MetadataVersion::Number(1),
                     &config.max_root_size,
                     config.min_bytes_per_second,
                     None,
-                )
+                )?;
+
+                local.store_metadata(&root_path, &MetadataVersion::Number(1), &root)?;
+
+                // FIXME: should we also the root as `MetadataVersion::None`?
+
+                Ok(root)
             })?;
 
         let tuf = Tuf::from_root_pinned(root, trusted_root_keys)?;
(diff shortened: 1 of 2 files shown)


src/client.rs
```diff
@@ -195,162 +203,161 @@ where
...
-            if !tuf.update_root(signed)? {
+
+            if !self.tuf.update_root(signed_root.clone())? {
                 error!("{}", err_msg);
                 return Err(Error::Programming(err_msg.into()));
             }
+
+            // FIXME what if this fails? do we need to roll back `self.tuf`s version?
+            self.local
+                .store_metadata(&root_path, &root_version, &signed_root)?;
```
heartsucker: I think we don't need to roll back the version number. The in-memory `Tuf` struct in memory is the absolute truth. The local repo is just for convenience and caching. We could issue a warning.

src/client.rs
```diff
@@ -195,162 +203,161 @@ where
...
         }
 
-        if !tuf.update_root(latest_root)? {
+        if !self.tuf.update_root(latest_root.clone())? {
             error!("{}", err_msg);
             return Err(Error::Programming(err_msg.into()));
         }
+
+        // FIXME what if this fails? do we need to roll back `self.tuf`s root?
+        self.local.store_metadata(
```
heartsucker: See above comment.

src/client.rs
```diff
@@ -195,162 +203,161 @@ where
...
+            &root_path,
+            &MetadataVersion::Number(latest_version),
+            &latest_root,
+        )?;
+
+        // FIXME should we also the root as `MetadataVersion::None`?
```
heartsucker: I think we need to store the latest as unversioned too since that would allow us in some cases to short cut and skip all the lower versions. If we don't the client might try to fetch it remotely thinking there is no local root at all. This isn't the current behavior, but this is a hedge against regressions.

src/client.rs
```diff
@@ -195,162 +203,161 @@ where
...
+            self.config.min_bytes_per_second,
             None,
         )?;
-        tuf.update_timestamp(ts)
+
+        if self.tuf.update_timestamp(signed_timestamp.clone())? {
+            let latest_version = signed_timestamp.version();
+
+            // FIXME what if this fails? do we need to roll back `self.tuf`s version?
+            self.local.store_metadata(
```
heartsucker: See above. No rollback.

src/client.rs
```diff
@@ -195,162 +203,161 @@ where
...
             &version,
             &Some(snapshot_description.size()),
-            config.min_bytes_per_second,
+            self.config.min_bytes_per_second,
             Some((alg, value.clone())),
         )?;
-        tuf.update_snapshot(snap)
+
+        if self.tuf.update_snapshot(signed_snapshot.clone())? {
+            // FIXME what if this fails? do we need to roll back `self.tuf`s version?
```
heartsucker: This is non-trivial. We'd have to actually keep all the TUF metadata in memory for all versions and then try to roll back to a consistent state of

1) Valid root chain
2) Valid time stamp pointing to:
3) A valid snapshot pointing to:
4) A valid targets (optionally pointing to valid delegations)

This is a huge ask. I think the simplest thing to do is abort and let the application that is implementing it try to start over.

src/client.rs
```diff
@@ -362,26 +369,37 @@ where
...
             &version,
             &Some(targets_description.size()),
-            config.min_bytes_per_second,
+            self.config.min_bytes_per_second,
             Some((alg, value.clone())),
         )?;
-        tuf.update_targets(targets)
+
+        if self.tuf.update_targets(signed_targets.clone())? {
+            // FIXME what if this fails? do we need to roll back `self.tuf`s version?
```
heartsucker: See above. Just keep going and log.

heartsucker (changes requested):
One small change.

erickt:
@heartsucker Updated the patch to make any errors saving metadata to the local repository just a warning. Hopefully that's what you wanted.

Also, I came up with another question. When we initially create a client and if we don't have metadata locally, should we download the first metadata, or the latest one? I think the latest one since who knows how old the first metadata is, but I wanted to be safe.

Or should we make this configurable? I imagine we would need to associate pinned keys with the root metadata they came from, because if we have years of metadata, we might want to update our pinned keys to something more recent than version 1's keys, since they could have been compromised.

trishankatdatadog:
@erickt I think it's always a good idea to ship clients with at least the latest root metadata file at the time of shipping. But if the client for some reason doesn't have anything, then it's best to download the latest root metadata file over TLS, but this implies TOFU: if the server has been compromised, then the client is in trouble.

trishankatdatadog:
Sorry, I hope you guys don't mind I jumped into the conversation :)

erickt:
@trishankatdatadog please do jump into these conversations! That's very helpful. I suppose if we really don't have some pinned keys to start from, it doesn't really matter what version we start from. I opened #162 to track this.

@heartsucker friendly ping :)

heartsucker (approved):

PR #158, opened 2018-08-15
Root serialization should put roles in it's own scope

erickt:
According to 4.4.3 of the spec, the root metadata serialization has the roles defined in their own object:

https://github.com/theupdateframework/specification/blob/master/tuf-spec.md#4-document-formats


heartsucker (changes requested):
Ok, so if we're shooting for compatibility with `go-tuf`, then we're also going to need to update the docs in `src/interchange/mod.rs` to correctly describe the metadata. Also, this is one of a number of changes we're going to need to make this work.

erickt:
Finally got around to updating the docs :)

heartsucker (approved):

PR #159, opened 2018-08-20
Simplify metadata construction with builders

erickt:
This patch series makes it a bit easier to construct metadata by wrapping the process with builders.

Additionally, to make it easier to work with generic metadata, it both moves the `version` and `expires` into the `Metadata` trait, and also rewrites the `SignedMetadata` to wrap a generic `Metadata` object. This allows us to more easily access the signed metadata state.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -259,7 +259,7 @@ where
             config.min_bytes_per_second,
             None,
         )?;
-        let latest_version = latest_root.signed().version();
+        let latest_version = latest_root.as_ref().version();
 
         if latest_version < tuf.root().version() {
             return Err(Error::VerificationFailure(format!(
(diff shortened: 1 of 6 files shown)


erickt:
Right now it draws inspiration for default metadata values from go-tuf, where it defaults to the version to be `1` and expires at:

* root - 365 days from now
* snapshot - 7 days from now
* targets - 90 days from now (edit: used to be 3 days)
* timestamp - 1 day from now

Are these the right values?

trishankatdatadog:
It depends on your use case, but I would imagine for typical scenarios, it should be:

* root - 365 days from now (offline keys, this is reasonable)
* targets - 1 month from now (offline keys, this seems reasonable)
* timestamp + snapshot (1 day, typically online keys, this is reasonable)

What do you guys think? @JustinCappos

erickt:
Oops, I misread go-tuf, it expires targets metadata after 3 months. I've updated the PR to use that.

trishankatdatadog:
That is extremely reasonable!

heartsucker (approved):

PR #151, opened 2018-08-01
Migrate over to futures 0.1 and hyper 0.12 (don't merge yet)

erickt:
This patch rewrites tuf to be asynchronous and concurrent, which requires a substantial amount of changes that frankly make some things a lot uglier. Async/Await however should really clean this up, but I wasn't sure if you would want to depend on an unstable feature that won't ship for at least a couple months.

I'm marking this as don't merge because I think there's still some more cleanup I can do to get it nicer, but I wanted to put it in front of you first to see what you think.

This patch depends on #139, and closes #98.

PR #166, opened 2018-12-11
Switch to futures, async/await, and hyper 0.12

erickt:
This is an experiment on porting rust-tuf over to async/await, which is what we use on Fuchsia. This is in contrast with #151, which uses futures 0.1, which is compatible with stable rust, but dramatically more invasive of a change, which is why I haven't rebased it on top of remotes/origin/develop yet.
    
The downside of this is that landing this patch would make rust-tuf require nightly until async/await is stablized. Unfortunately there is no clear date on when that is going to occur.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -3,10 +3,12 @@
 //! # Example
 //!
 //! ```no_run
+//! #![feature(async_await, await_macro, futures_api, pin)]
+//! # use futures::executor::block_on;
 //! # use hyper::client::Client as HttpClient;
 //! # use hyper::Url;
 //! # use std::path::PathBuf;
-//! # use tuf::Tuf;
+//! # use tuf::{Result, Tuf};
 //! # use tuf::crypto::KeyId;
 //! # use tuf::client::{Client, Config};
 //! # use tuf::metadata::{RootMetadata, SignedMetadata, Role, MetadataPath,
@@ -19,13 +21,13 @@
 //!     "T5vfRrM1iHpgzGwAHe7MbJH_7r4chkOAphV3OPCCv0I=",
 //! ];
 //!
-//! # fn main() {
+//! # fn main() -> Result<()> {
+//! # block_on(async {
 //! let key_ids: Vec<KeyId> = TRUSTED_ROOT_KEY_IDS.iter()
 //!     .map(|k| KeyId::from_string(k).unwrap())
 //!     .collect();
 //!
-//! let local = FileSystemRepository::<Json>::new(PathBuf::from("~/.rustup"))
-//!     .unwrap();
+//! let local = FileSystemRepository::<Json>::new(PathBuf::from("~/.rustup"))?;
 //!
 //! let remote = HttpRepository::new(
 //!     Url::parse("https://static.rust-lang.org/").unwrap(),
@@ -33,30 +35,32 @@
 //!     Some("rustup/1.4.0".into()),
 //!     None);
 //!
-//! let mut client = Client::with_root_pinned(
+//! let mut client = await!(Client::with_root_pinned(
 //!     &key_ids,
 //!     Config::default(),
 //!     local,
 //!     remote,
-//! ).unwrap();
-//! let _ = client.update().unwrap();
+//! ))?;
+//!
+//! let _ = await!(client.update())?;
+//! # Ok(())
+//! # })
 //! # }
 //! ```
 
 use chrono::offset::Utc;
+use futures::io::{AsyncRead, AsyncReadExt, AsyncWrite};
 use log::{error, warn};
-use std::io::{Read, Write};
 
 use crate::crypto::{self, KeyId};
 use crate::error::Error;
 use crate::interchange::DataInterchange;
 use crate::metadata::{
-    Metadata, MetadataPath, MetadataVersion, Role, RootMetadata, SignedMetadata, SnapshotMetadata,
+    Metadata, MetadataPath, MetadataVersion, Role, SignedMetadata, SnapshotMetadata,
     TargetDescription, TargetPath, TargetsMetadata, VirtualTargetPath,
 };
 use crate::repository::Repository;
 use crate::tuf::Tuf;
-use crate::util::SafeReader;
 use crate::Result;
 
 /// Translates real paths (where a file is stored) into virtual paths (how it is addressed in TUF)
@@ -67,9 +71,9 @@ use crate::Result;
 /// ```
 /// # use tuf::client::{PathTranslator, DefaultTranslator};
 /// # use tuf::metadata::{VirtualTargetPath, TargetPath};
-/// # let path = TargetPath::new("foo".into()).unwrap();
-/// # let virt = VirtualTargetPath::new("foo".into()).unwrap();
-/// # let translator = DefaultTranslator::new();
+/// let path = TargetPath::new("foo".into()).unwrap();
+/// let virt = VirtualTargetPath::new("foo".into()).unwrap();
+/// let translator = DefaultTranslator::new();
 /// assert_eq!(path,
 ///            translator.virtual_to_real(&translator.real_to_virtual(&path).unwrap()).unwrap());
 /// assert_eq!(virt,
@@ -121,23 +125,26 @@ where
 impl<D, L, R, T> Client<D, L, R, T>
 where
     D: DataInterchange,
-    L: Repository<D>,
-    R: Repository<D>,
-    T: PathTranslator,
+    L: Repository<D> + 'static,
+    R: Repository<D> + 'static,
+    T: PathTranslator + 'static,
 {
     /// Create a new TUF client. It will attempt to load initial root metadata from the local repo
     /// and return an error if it cannot do so.
     ///
     /// **WARNING**: This method offers weaker security guarantees than the related method
     /// `with_root_pinned`.
-    pub fn new(config: Config<T>, local: L, remote: R) -> Result<Self> {
-        let root = local.fetch_metadata(
-            &MetadataPath::from_role(&Role::Root),
-            &MetadataVersion::Number(1),
+    pub async fn new(config: Config<T>, local: L, remote: R) -> Result<Self> {
+        let root_path = MetadataPath::from_role(&Role::Root);
+        let root_version = MetadataVersion::Number(1);
+
+        let root = await!(local.fetch_metadata(
+            &root_path,
+            &root_version,
             &config.max_root_size,
             config.min_bytes_per_second,
             None,
-        )?;
+        ))?;
 
         let tuf = Tuf::from_root(root)?;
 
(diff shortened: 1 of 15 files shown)


src/client.rs
```diff
@@ -3,63 +3,60 @@
...
 //!
-//!     let mut client = Client::with_root_pinned(
-//!         &key_ids,
-//!         Config::default(),
-//!         local,
-//!         remote,
-//!     ).unwrap();
-//!     let _ = client.update().unwrap();
-//! }
+//! let mut client = block_on(Client::with_root_pinned(
```
cramertj: I wouldn't use `block_on` in examples since it won't work with many reactor/executor setups. instead, consider following the pattern we use in the `futures` crate and wrapping the code in an `async` block and `await!`ing. If you want it to run as part of doctests, you can wrap it in a hidden `block_on` invocation.

src/client.rs
```diff
@@ -156,42 +156,40 @@ where
     /// repositories using the provided key IDs to pin the verification.
     ///
     /// This is the preferred method of creating a client.
-    pub fn with_root_pinned<'a, I>(
-        trusted_root_keys: I,
+    pub async fn with_root_pinned(
+        trusted_root_keys: &[KeyId],
```
cramertj, line 164: Why did the type here change?

src/client.rs
```diff
@@ -206,26 +204,26 @@ where
     /// Update TUF metadata from the remote repository.
     ///
     /// Returns `true` if an update occurred and `false` otherwise.
-    pub fn update(&mut self) -> Result<bool> {
-        let r = self.update_root()?;
-        let ts = self.update_timestamp()?;
-        let sn = self.update_snapshot()?;
-        let ta = self.update_targets()?;
+    pub async fn update(&mut self) -> Result<bool> {
+        let r = await!(self.update_root())?;
```
cramertj, line 212: This will execute each of these updates in series-- is that what you want, or should the updates be able to run concurrently?

src/client.rs
```diff
@@ -268,33 +266,31 @@ where
...
 
-        self.store_metadata(
-            &root_path,
-            &MetadataVersion::Number(latest_version),
-            &latest_root,
-        );
-        self.store_metadata(&root_path, &MetadataVersion::None, &latest_root);
+        let latest_version = MetadataVersion::Number(latest_version);
+
+        await!(self.store_metadata(&root_path, &latest_version, &latest_root,));
```
cramertj, line 296: ditto RE "can these happen concurrently"?

src/client.rs
```diff
@@ -401,17 +395,18 @@ where
         };
 
         let targets_path = MetadataPath::from_role(&Role::Targets);
+        let targets_size = Some(targets_description.size());
```
cramertj, line 402: huh, it's unfortunate that these temporaries aren't living long enough (guessing that's why you made this change). There are merged-but-unimplemented RFCs that would fix this, but it's still disappointing.

src/repository.rs
```diff
@@ -117,136 +125,155 @@ impl<D> Repository<D> for FileSystemRepository<D>
...
+                path.extend(meta_path.components::<D>(&version));
 
-        let read = SafeReader::new(
-            File::open(&path)?,
-            max_size.unwrap_or(::std::usize::MAX) as u64,
-            min_bytes_per_second,
-            hash_data,
-        )?;
+                let mut reader = SafeReader::new(
+                    AllowStdIo::new(File::open(&path)?),
```
cramertj, line 177: This will block the event loop on doing file IO. I'd recommend using a dedicated thread for file IO. The same goes for all other usages here and below of `AllowStdIo`.

src/repository.rs
```diff
@@ -117,136 +125,155 @@ impl<D> Repository<D> for FileSystemRepository<D>
...
-            return Err(Error::NotFound);
-        }
+                if !path.exists() {
+                    return Err(Error::NotFound);
+                }
 
-        let (alg, value) = crypto::hash_preference(target_description.hashes())?;
+                let (alg, value) = crypto::hash_preference(target_description.hashes())?;
+
+                let reader: Box<dyn AsyncRead> = Box::new(SafeReader::new(
```
cramertj, line 234: Why change this from `SafeReader` directly to `Box`ing it?

src/repository.rs
```diff
@@ -312,83 +349,115 @@ where
...
-            resp,
-            max_size.unwrap_or(::std::usize::MAX) as u64,
-            min_bytes_per_second,
-            hash_data,
-        )?;
-        Ok(D::from_reader(read)?)
+                let mut buf = Vec::new();
+                await!(reader.read_to_end(&mut buf))?;
+
+                Ok(D::from_reader(buf.as_slice())?)
```
cramertj: This may be a larger change to save for the future, but should the `DataInterchange` trait be changed to work over `AsyncRead`/`AsyncWrite` rather than `Read`/`Write` so as to prevent the intermediate buffer here?

src/repository.rs
```diff
@@ -314,16 +322,25 @@ where
...
+        let uri: Uri = url.into_string().parse().map_err(|_| {
+            Error::IllegalArgument(format!("URL was 'cannot-be-a-base': {:?}", self.url))
+        })?;
+
+        let req = Request::builder()
+            .uri(uri)
+            .header("User-Agent", &*self.user_agent)
+            .body(Body::default())?;
+
+        let resp = await!(Compat01As03::new(self.client.request(req)))?;
```
cramertj: nit: you can spell this as just `.compat()`

src/client.rs
erickt: I was getting the following error:

```rust
error[E0700]: hidden type for `impl Trait` captures lifetime that does not appear in bounds
   --> src/client.rs:168:10
    |
168 |     ) -> Result<Self>
    |          ^^^^^^^^^^^^
    |
note: hidden type `impl core::future::future::Future` captures the lifetime 'a as defined on the method body at 163:35
   --> src/client.rs:163:35
    |
163 |     pub async fn with_root_pinned<'a, I>(
    |                                   ^^
```

Wasn't sure if this was a limitation of impl Future/async/await, or it's actually supposed to work and it's a compiler error. Do you have a sense?

src/client.rs
erickt: Yeah, this serialization is mandated by the [tuf spec](https://github.com/theupdateframework/specification/blob/master/tuf-spec.md#5-detailed-workflows). Each metadata depends on the last. The methods internally will skip actually downloading if it's not necessary.

src/client.rs
erickt: These however could be in parallel, though in practice this will just be writing to the local filesystem. Is there much advantage of still making this parallel?

src/client.rs
erickt: Yeah it was a little unfortunate, but as you see it wasn't too painful to implement, so it's not that bad.

src/repository.rs
erickt: Oh drat. Do you know of a good example of someone spawning a `AsyncRead` impl on a separate thread, or have a suggestion on how to implement it? I couldn't find any on github. I did find [futures-fs](https://github.com/seanmonstar/futures-fs), but it's written to be a `Stream`, and it's written against 0.1.

src/repository.rs
erickt: Do you mean changing this into:

```rust
let reader = SafeReader::new(...);
Ok(Box::new(reader) as Box<dyn AsyncRead>)
```

?, if so, yeah I could switch to that. Or are you referring to hiding the fact that we're using an `AsyncRead`? If the latter, I'm a little back and forth on it. The `FilesystemRepository` doesn't really need the full `SafeReader` interface, it really just needs a reader that asserts the hash is correct. That suggested to me that we might want to break up the `SafeReader`, assuming there's a performance benefit. But we lose guaranteeing that other impls of `Repository` have to protect against the things `SafeReader` protects against. I go back and forth.

@heartsucker: do you have an opinion on this?

src/repository.rs
erickt: We could, but `DataInterchange` really is just doing a serde_json deserialization, that only supports deserializing from `&[u8]` and `std::io::Read`. I can however at least add an api for `DataInterchange::from_slice`, which may be a little faster.

src/client.rs
cramertj: This is supposed to work, but currently does not.

src/repository.rs
cramertj: I don't have a suggestion, unfortunately-- it really is just a matter of having to spin up a separate thread and do the work there.

src/repository.rs
heartsucker: This was originally done for simplicity and for being able to reuse the `Repository` type as both local and remote. It should probably be two types because local repositories can be semi-trusted. Whatever gets this merged in is fine by me.

src/repository.rs
heartsucker: Extending the API should be fine. It might also make sense to strip out the `DataInterchage` type completely. I was trying to make this lib a bit more generalizable, but since it seems no one out there is doing TUF with anything but JSON, this might just add unnecessary complexity. If we want to remove it, we could open a ticket for that. This has been listed as beta in the readme this whole time, so if we broke compatibility, I would consider that ok. Also it's going from `0.2.x` to `0.3.x` so semver says that's ok too.

heartsucker (commented):
Since the develop branch right now seems principally targeted towards Fuschia, I think it's ok to move this to nightly. I haven't had time to keep up to date on the latest futures/async news so I can't really form a strong opinion either way right now.

src/repository.rs
erickt: There's actually a new [TAP](https://github.com/theupdateframework/taps/pull/106) filed last month that adds support for multiple transport formats.

We also might want to explore other formats for optimizing storage and transport overhead, so I think it's worth keeping around right now.

erickt:
@heartsucker great! I think the big thing to address is that there's a performance hit using `AllowStdIo` in the repository, which I'd like to try to address in a future patch. I want to do some research into this space, and on splitting repositories into a `LocalRepsitory` and `RemoteRepository` to see if that'd help. If this patch looks good to you, would you want to merge it in?

src/repository.rs
heartsucker: Ok so are you still shooting for compatibility with DockerHub or Notary or whatever it was?

heartsucker:
@erickt This is going to be an improvement over the current setup which is fully synchronous, so I think it's safe to merge. I need to review in depth just for protocol / security stuff, but generally yes I'm fine with merging this.

src/repository.rs
```diff
@@ -117,136 +125,155 @@ impl<D> Repository<D> for FileSystemRepository<D>
...
 /// A repository accessible over HTTP.
-pub struct HttpRepository<D>
+pub struct HttpRepository<C, D>
 where
+    C: Connect + Sync,
     D: DataInterchange,
 {
     url: Url,
-    client: Client,
+    client: Client<C>,
```
heartsucker: What was your reasoning for moving the client into the repository? My original API design was to have the client be the primary entity a user interacts with.

PR #175, opened 2019-01-18
Update ring

vks:


erickt:
Thanks for this! I was hoping to wait to land this for rustls 0.15 to come out which adds support for ring 0.14, since you can't have multiple versions of ring in the same binary :(

erickt:
Thanks @vks!

PR #176, opened 2019-02-28
Use "length" instead of "size" to match the spec

erickt:
According to the [TUF spec](https://github.com/theupdateframework/specification/blob/master/tuf-spec.md), metadata uses "length" instead of "size" in the metadata. This patch switches over to using that naming scheme.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -140,7 +140,7 @@ where
         let root_version = MetadataVersion::Number(1);
 
         let root =
-            await!(local.fetch_metadata(&root_path, &root_version, &config.max_root_size, None))?;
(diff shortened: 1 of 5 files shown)


heartsucker (commented):
This would also need to update the docs notes in `src/interchange/mod.rs`. Since the spec is ever advancing, and we are aiming for compatibility with the go-tuf impl, we will need to maintain our own version of the spec as a reference.

erickt:
Updated the doc! Hopefully this version will get through appveyor. Thanks as always @heartsucker!

heartsucker (approved):

PR #191, opened 2019-08-01
Allow {FileSystem,Http}Repository to customize metadata/targets prefix

### LINKED ISSUE #182: Customize {FileSystem,Http}Repository directory structure
FileSystemRepository assumes that metadata is stored under `metadata/` and targets are stored under `targets/`. HttpRepository allows for metadata to be customized, but assumes targets are found at the top level. We should allow this to be customized since some projects may structure their repositories differently (for example, go-tuf puts their metadata in the top level directory).

erickt:
This allows users to target repositories with alternative layouts.

Closes #182

### DIFF
--- a/src/repository.rs
+++ b/src/repository.rs
@@ -114,7 +114,7 @@ where
         &'a self,
         meta_path: &'a MetadataPath,
         version: &'a MetadataVersion,
-        max_length: &'a Option<usize>,
+        max_length: Option<usize>,
         hash_data: Option<(&'static HashAlgorithm, HashValue)>,
     ) -> BoxFuture<'a, Result<SignedMetadata<D, M>>>
     where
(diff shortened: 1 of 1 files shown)


src/repository.rs
```diff
@@ -92,12 +92,74 @@ where
     }
 }
 
+
+/// A builde rto create a repository contained on the local file system.
```
heartsucker: Typo: `builde rto` -> `builder to`

heartsucker (commented):
Minor need. Needs a rebase too.

src/repository.rs
erickt: fixed, thanks

erickt:
Rebased (including being on top of #192). Thanks for all the reviews!

heartsucker (approved):

PR #197, opened 2019-08-01
 Include spec_version in role metadata

erickt:
In the [spec] in section 4, each metadata role includes a "spec_version"
field that corresponds to the TUF metadata format.

### DIFF
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -1900,6 +1900,7 @@ mod test {
 
         let jsn = json!({
             "_type": "root",
+            "spec_version": "1.0",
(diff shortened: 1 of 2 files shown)


mnm678:
This looks good, you may want to look at an in progress TAP about [TUF version management]( https://github.com/theupdateframework/taps/pull/107). It is trying to allow for updates to the specification version to allow for breaking changes. This is still ongoing, but any feedback would be appreciated.

erickt:
Rebased on top of master

@mnm678 and thanks! I'll check out that spec.

PR #202, opened 2019-08-02
lift block_on in client::test::root_chain_update to outer scope

erickt:


erickt:
Rebased on top of master

heartsucker (approved):

PR #203, opened 2019-08-02
migrate key management to be TUF 1.0 compliant

### LINKED ISSUE #150: Changing key id to be hexdigest(sha256(cjson(public_key)))
Fuchsia' tuf implementation needs to be compatible with [go-tuf](https://github.com/flynn/go-tuf), which uses `hexdigest(sha256(cjson(public_key)))` for it's key id, which is currently required by the [spec section 4.2](https://github.com/theupdateframework/specification/blob/master/tuf-spec.md#4-document-formats). Would you be okay if I converted rust-tuf to be compliant by default, at least until [TAP 11](https://github.com/theupdateframework/taps/pull/74) is accepted? I could migrate the current key id algorithm behind a feature flag (or put the standard behind a feature flag, if you'd prefer that).

### LINKED ISSUE #198: key type field is "keytype", value is "keyval" object
In the [spec] in section 4.2, keys store their type in a field `"keytype"`, and the value in `"keyval": {"public": <key>}` object.

erickt:
This updates rust-tuf to be compliant with how TUF 1.0 [spec] encodes keys and key material.

* In section 4.2 defines that key ids are calculated using hexdigit(sha256(cjson(public_key))).
* In section 4.2, keys store their type in a field "keytype", and the value in "keyval": {"public": } object.
* In section 4.2, ed25519 public keys should be stored as the hexdigest(key_bytes).

Note this adds the "pretty_assertions" dev dependency, which highlights differences between comparisons.

Closes #150
Closes #198


src/crypto.rs
```diff
@@ -110,67 +110,61 @@ pub fn calculate_hashes<R: Read>(
...
+/// A `KeyId` is calculated as the hex digest of the SHA-256 hash of the canonical form of the
+/// public key, or `hexdigest(sha256(cjson(public_key)))`.
 #[derive(Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
-pub struct KeyId(Vec<u8>);
+pub struct KeyId(String);
 
 impl KeyId {
-    /// Parse a key ID from a base64url string.
+    /// Parse a key ID from a string.
     pub fn from_string(string: &str) -> Result<Self> {
```
ComputerDruid: impl FromStr instead?

src/crypto.rs
```diff
@@ -545,8 +538,20 @@ impl PublicKey {
...
         })?;
 
-        let key_id = calculate_key_id(der_bytes);
-        Ok(PublicKey { typ, key_id, scheme, value: PublicKeyValue(value) })
+        Self::new(typ, scheme, PublicKeyValue(value))
+    }
+
+    fn from_ed25519(bytes: Vec<u8>) -> Result<Self> {
+        if bytes.len() != 32 {
+            return Err(Error::IllegalArgument("ed25519 keys must be 44 bytes long".into()));
```
ComputerDruid: s/44/32/

src/crypto.rs
```diff
@@ -545,8 +538,20 @@ impl PublicKey {
...
+        }
+
+        Ok(PublicKey {
```
ComputerDruid: You use `PublicKey::new()` in `from_spki()`, you should probably do the same thing in both places.

erickt:
Merged in the latest remotes/origin/upstream/develop

heartsucker (approved):

PR #207, opened 2019-08-06
Sync Timestamp and Snapshot metadata with TUF 1.0 schema

### LINKED ISSUE #193: Timestamp and Snapshot meta field stores role name with extension
In the [spec](https://github.com/theupdateframework/specification/blob/master/tuf-spec.md), the METAFILES object defined in section 4.4 states that the metadata path includes the suffix, such as `"snapshot.json"`. rust-tuf, however is only storing `"snapshot"` in the generated files.

erickt:
This fixes a couple inconsistencies between rust-tuf and the [spec]:

* In section 4.6, the timestamp metadata stores the reference to the snapshot in a "meta" object, just like the snapshot metadata.
* In section 4.4 and 4.6, the snapshot and timestamp stores metadata with a ".json" suffix. This patch unfortunately hardcodes the extension, which makes it unportable across interchanges, but since we don't support multiple interchange formats yet, this is more of a theoretical problem that can be addressed in the future.
* Snapshot now fails to validate if there are duplicate metadata fields.

In addition, fixes a typo in a root metadata test.

Closes #193


src/metadata.rs
```diff
@@ -953,15 +956,15 @@ impl TimestampMetadataBuilder {
 pub struct TimestampMetadata {
     version: u32,
     expires: DateTime<Utc>,
-    snapshot: MetadataDescription,
+    meta: HashMap<MetadataPath, MetadataDescription>,
```
ComputerDruid: By implementing it this way, you accept timestamp.json files which contain roles other than snapshot.json to be valid. I can't think of why you'd want to put another role in the timestamp, and the workflow doesn't use that information. Would it make sense to only allow snapshot in timestamp?

The TUF spec isn't really clear on this. What do python-tuf and go-tuf do?

trishankatdatadog: Well, technically, I think you can list anything you like in the timestamp other than metadata about snapshot, so long as you ignore them, although it should be a warning flag

So it's safe to say that for all practical purposes, anything other than metadata about snapshot in the timestamp metadata should be rejected

What do you think, @JustinCappos?

JustinCappos:
I'd recommend only allowing snapshot.  It seems bad to allow other things
to be listed that we did not expect.

erickt:
Great, I'll update this PR to only allow the timestamp role to reference the snapshot role. I also filed https://github.com/theupdateframework/specification/pull/49 to update the spec to:

> METAFILES is the same is described for the snapshot.json file. In the case of the timestamp.json file, this MUST only include a description of the snapshot.json file.

erickt:
Updated to require that the timestamp metadata "meta" field only contains "snapshot.json".

src/interchange/cjson/shims.rs
```diff
@@ -140,14 +147,14 @@ impl TimestampMetadata {
...
             expires: format_datetime(metadata.expires()),
-            snapshot: metadata.snapshot().clone(),
+            meta: TimestampMeta { snapshot: metadata.snapshot().clone() },
         })
     }
 
     pub fn try_into(self) -> Result<metadata::TimestampMetadata> {
         if self.typ != metadata::Role::Timestamp {
             return Err(Error::Encoding(format!(
-                "Attempted to decode datetime metdata labeled as {:?}",
```
heartsucker, line 150: lmao oh `sed`, sometimes you do us dirty.

heartsucker (approved):

PR #208, opened 2019-08-06
Remove support for unknown KeyTypes and SignatureSchemes

erickt:
The only thing we do with unknown KeyType and SignatureSchemes is report and error if we attempt to use them. So instead, we should just not represent them in the crypto module.

### DIFF
--- a/src/crypto.rs
+++ b/src/crypto.rs
@@ -5,10 +5,7 @@ use derp::{self, Der, Tag};
 use ring;
 use ring::digest::{self, SHA256, SHA512};
 use ring::rand::SystemRandom;
-use ring::signature::{
-    Ed25519KeyPair, KeyPair, RsaKeyPair, ED25519, RSA_PSS_2048_8192_SHA256,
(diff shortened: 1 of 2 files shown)


heartsucker (commented):
When dealing with crypto things, I prefer verbose logging of error messages. I find it infuriating when I get something like "signature invalid" with no indication why. I really think they should stay in to help people figure out why things went wrong, especially if client/sever are running different versions of the protocol.

erickt:
@heartsucker this PR still has semantic errors, we just get them now when we deserialize the metadata, instead of when we try to verify the metadata signature. I just pushed up a new version that includes some tests to assert the particular error we get during deserialization. Does that address your concern?

ComputerDruid (approved):

heartsucker:
> we just get them now when we deserialize the metadata

To me this is incorrect behavior. It should be allowed to serialize unknown keys/signature types into the metadata for forward compatibility. Also, this would make the error happen during `tuf update` and not `tuf install foo` which is where I would expect such an error to arise.

erickt:
Yep you are right. Closing this ticket.

PR #210, opened 2019-08-08
Add support for custom TargetDescription metadata

erickt:
This extends TargetDescription to capture the `custom` field as a map of `String`s to `serde_json::Value`s. This ties the interchange more closely with the cjson format, but this could be addressed once more interchange formats are implemented.

In addition, this adds a new function `Client::fetch_target_description`, to allow end users to discover this target information.


src/metadata.rs
```diff
@@ -1376,23 +1376,28 @@ impl TargetPath {
 }
 
 /// Description of a target, used in verification.
-#[derive(Debug, Clone, PartialEq, Serialize)]
+#[derive(Debug, Clone, PartialEq)]
 pub struct TargetDescription {
     length: u64,
     hashes: HashMap<HashAlgorithm, HashValue>,
+    custom: HashMap<String, serde_json::Value>,
```
heartsucker: Shouldn't this be optional? So far as I recall, it's optional in the spec even if it's always there in `go-tuf`.

trishankatdatadog: Yes, it should be optional

erickt: I can make it explicitly optional. I this up such that we won't add `"custom"` if the `custom` map is empty, but making this field an `Option<HashMap<String, serde_json::Value>>` would be more semantically correct.

erickt: Actually, I've changed my mind :) I can't think of a case where I'd want to do something different if some target has an empty `"custom"` map compared to that target having no `"custom"` field. The shim in this patch serializes the custom field only if it's empty, otherwise it ignores it, and deserializes to an empty map if it's not present in the metadata. Can anyone come up with a counter example where this pattern is wrong?

trishankatdatadog: Well, technically, consumers are free to ignore custom targets metadata, if they so wish, so this should be ok.

ComputerDruid: Won't this cause a problem for verifying metadata that contains `custom: {}` ?

heartsucker: I think it is an important difference to differentiate between absent, null, and empty custom. Because we're fundamentally dealing with JSON right now and those are all different in JSON land. This library should be able to support creating and signing metadata, and that's not possible if these three cases are all treated as identical.

erickt: good catch! Yeah that would cause the signatures to be different because we aren't internally saving the raw unparsed json value. I'll switch back to `Option<HashMap<String, Value>>`.

erickt:
Updated the PR to capture the custom metadata in a `Option<HashMap<String, Value>>`, so that we can re-compute the signature if a target has an empty "custom" field.

erickt:
Note that as far as I can tell, the `"custom"` is supposed to be a map, which is how python-tuf [implements it](https://github.com/theupdateframework/tuf/blob/4f01a31346c2da9c0fb79a62cd6bb77da48ba5ed/tuf/formats.py#L231). I filed https://github.com/theupdateframework/specification/issues/50 for clarification, but I think we can keep this field represented as a `Option<HashMap>`.

heartsucker:
> Updated the PR to capture the custom metadata in a Option<HashMap<String, Value>>, so that we can re-compute the signature if a target has an empty "custom" field.

At some point if we introduce signing, we will have to figure out a way to reject `null` values. A problem with this is that a lot of this would tie the library to JSON. Because `null` and absent are equivalent according to `serde`.

PR #212, opened 2019-08-29
 Include key type and scheme in PublicKey's PartialEq and Hash

erickt:
Now that we've switched over to using the TUF 1.0 scheme for keyids, the public key's value no longer incorporates the key's type and signature scheme. This fixes things so that equality and hashing work correctly with different key type and signature schemes, but the same public key bytes.

### DIFF
--- a/src/crypto.rs
+++ b/src/crypto.rs
@@ -180,7 +180,7 @@ impl<'de> Deserialize<'de> for KeyId {
 }
 
 /// Cryptographic signature schemes.
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
 pub enum SignatureScheme {
     /// [Ed25519](https://ed25519.cr.yp.to/)
     #[serde(rename = "ed25519")]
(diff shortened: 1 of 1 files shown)


src/crypto.rs
```diff
@@ -595,8 +595,7 @@ impl PublicKey {
 
 impl PartialEq for PublicKey {
     fn eq(&self, other: &Self) -> bool {
-        // the other two fields are derived from this one, we ignore them
```
ComputerDruid, line 598: Maybe leave this comment explaining that key_id is computed from these 3?

src/interchange/cjson/shims.rs
```diff
@@ -381,7 +381,9 @@ impl TargetDescription {
         TargetDescription {
             length: description.length(),
             hashes: description.hashes().iter().map(|(k, v)| (k.clone(), v.clone())).collect(),
-            custom: description.custom().map(|custom| custom.iter().map(|(k, v)| (k.clone(), v.clone())).collect()),
+            custom: description
+                .custom()
+                .map(|custom| custom.iter().map(|(k, v)| (k.clone(), v.clone())).collect()),
```
ComputerDruid: this looks unrelated to this change.

src/crypto.rs
```diff
@@ -616,7 +615,8 @@ impl PartialOrd for PublicKey {
 
 impl hash::Hash for PublicKey {
     fn hash<H: hash::Hasher>(&self, state: &mut H) {
-        // the other two fields are derived from this one, we ignore them
```
ComputerDruid, line 619: ditto about ignoring keyid

wellsie1116 (approved):

PR #213, opened 2019-09-04
Add go-tuf interop test suite

erickt:
We want to make sure that rust-tuf is compatible with go-tuf, and other TUF implementations. This patch starts that process by importing the [go-tuf] test metadata into the rust-tuf repository in order to make sure rust-tuf can parse the metadata. There are two sets of metadata:
    
* go-tuf: what the library currently generates), and
* go-tuf-transition-M4: what the library will eventually generate once all the patches have landed).

Unfortunately due to a few bugs (#224, #225, and #226), we can't correctly parse all the metadata yet, so this patch just ignores that output for now.

I recommend reviewing each patch individually, which might be easier to review.


tests/interop.rs
```diff
@@ -0,0 +1,156 @@
...
+    test_client(
+        Path::new("tests")
+            .join("interop")
+            .join("go-tuf-transition-M4")
+            .join("consistent-snapshot-false"),
+    );
+}
+
+#[test]
+#[ignore] // FIXME: Blocked roin #224, #225, #226
```
ComputerDruid: s/roin/on/

tests/interop.rs
```diff
@@ -0,0 +1,156 @@
+use futures_executor::block_on;
```
ComputerDruid: https://doc.rust-lang.org/cargo/reference/manifest.html#the-project-layout suggests this file could be named `tests/interop/main.rs` which would group the testdata with the test better.

tests/interop.rs
```diff
@@ -0,0 +1,156 @@
...
+    block_on(async {
+        let mut test = TestClient::new(dir).unwrap();
+        test.run().await;
+    })
+}
+
+struct TestClient {
+    test_steps: Vec<PathBuf>,
+    local: EphemeralRepository<Json>,
+    targets: BTreeMap<TargetPath, String>,
```
ComputerDruid: can you rename this expected_targets?

tests/interop.rs
```diff
@@ -0,0 +1,156 @@
...
+}
+
+fn test_client(dir: PathBuf) {
+    block_on(async {
+        let mut test = TestClient::new(dir).unwrap();
+        test.run().await;
+    })
+}
+
+struct TestClient {
```
ComputerDruid: Somewhere either here or in run or run_step, can you describe that this test "runs a local repository against a sequence of remote repositories" as a doc comment? It might also be worth documenting what the fields here are used for.

tests/interop.rs
```diff
@@ -0,0 +1,156 @@
...
+            self.run_step(&key_ids, step_dir).await;
+        }
+    }
+
+    async fn run_step(&mut self, key_ids: &[KeyId], dir: PathBuf) {
+        let remote = init_remote(&dir).unwrap();
+
+        // Connect to the client with our initial keys.
+        let mut client =
+            Client::with_root_pinned(&key_ids, Config::default(), &self.local, remote, 1)
```
ComputerDruid: Maybe off-topic, but shouldn't this specify a public key, not the key ID?

ComputerDruid (changes requested):

tests/interop.rs
erickt: You are right! we definitely should fix this. I filed https://github.com/heartsucker/rust-tuf/issues/229 to track this.

tests/interop/go-tuf/generate.go
```diff
@@ -0,0 +1,19 @@
+package main
```
ComputerDruid: I don't think this will build because it'll pull the wrong go-tuf. Instead of leaving this here, maybe point to the commit from upstream that these were taken from? It makes more sense to generate locally and then copy out anyway IMO.

Also, that implies that you should probably call this directory M3, since nothing will keep it in sync with go-tuf's master; all of these directories are snapshots.

tests/interop/go-tuf-transition-M4/generate.go
```diff
@@ -0,0 +1,146 @@
+package main
```
ComputerDruid: Instead of having this file, I think it'd make more sense to leave a README here explaining where this metadata came from

ComputerDruid (approved):
I just have a couple of comments about the metadata organization, the rest of this seems good.

tests/interop/go-tuf/generate.go
erickt: Good idea.

Regarding the name, I think we should also link it to our fork, so I prefixed the directory with `fuchsia-`. I don't want to use M3 though, since this metadata isn't what we called M3. I instead named this directory `fuchsia-go-tuf-HEAD`, and called out the specific commit in the README. That sound good?

tests/interop/go-tuf-transition-M4/generate.go
erickt: Likewise, I removed the go file, added README, and renamed this `fuchsia-go-tuf-transition-M4`.

erickt:
Updated with the following changes:
    
    * added READMEs to point at where the metadata came from.
    * renamed `go-tuf` metadata to `fuchsia-go-tuf-HEAD`
    * renamed `go-tuf-transition-M4` metadata to `fuchsia-go-tuf-transition-M4`

PR #217, opened 2019-11-05
Allow MestadataPath::new to take &strs

erickt:
This simplifies interacting with MetadataPaths when we have fixed strings.


tarcieri:
Mestadata? 😉

heartsucker (approved):

PR #219, opened 2019-11-06
Add rustfmt.toml; reformat with rustfmt 1.4.9-nightly

erickt:
This makes sure that rust-tuf is formatted with a consistent style, rather than potentially picking up a rustfmt.toml from a parent directory.


src/repository.rs
```diff
@@ -501,7 +505,10 @@ where
             if status == StatusCode::NOT_FOUND {
                 Err(Error::NotFound)
             } else {
-                Err(Error::Opaque(format!("Error getting {:?}: {:?}", self.url, resp)))
+                Err(Error::Opaque(format!(
+                    "Error getting {:?}: {:?}",
+                    self.url, resp
+                )))
```
ComputerDruid, line 511: This change is the worst of it, and it's not so bad.

ComputerDruid (approved):

PR #221, opened 2019-11-08
Add support for optional keyid_hash_algorithms in keys

kellerb:
The TUF 1.0 keyid algorithm is `hexdigit(sha256(cjson(public_key)))`.
python-tuf has an extension to the spec where keys in the root metadata
contain a `keyid_hash_algorithms` which can specify alternative algorithms
to sha256. In order to maintain compatibility with python-tuf and go-tuf,
this patch adds the ability to parse metadata that contains this field.
It also generates metadata with `keyid_hash_algorithms=["sha256"]`, since
python-tuf currently has a hard requirement that this field must be
present in the root metadata. However, this field is optional in rust-tuf,
and can parse metadata that doesn't contain the field.

### DIFF
--- a/src/crypto.rs
+++ b/src/crypto.rs
@@ -34,6 +34,10 @@ const RSA_SPKI_OID: &[u8] = &[0x2a, 0x86, 0x48, 0x86, 0xf7, 0x0d, 0x01, 0x01, 0x
 /// 1.3.101.112 curveEd25519(EdDSA 25519 signature algorithm)
 const ED25519_SPKI_OID: &[u8] = &[0x2b, 0x65, 0x70];
 
+fn python_tuf_compatibility_keyid_hash_algorithms() -> Option<Vec<String>> {
+    Some(vec!["sha256".to_string(), "sha512".to_string()])
+}
+
 /// Given a map of hash algorithms and their values, get the prefered algorithm and the hash
 /// calculated by it. Returns an `Err` if there is no match.
 ///
@@ -118,6 +122,7 @@ pub fn calculate_hashes<R: Read>(
 fn shim_public_key(
     key_type: &KeyType,
     signature_scheme: &SignatureScheme,
+    keyid_hash_algorithms: &Option<Vec<String>>,
     public_key: &[u8],
 ) -> ::std::result::Result<shims::PublicKey, derp::Error> {
     let key = match key_type {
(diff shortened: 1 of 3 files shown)


src/crypto.rs
```diff
@@ -627,7 +675,10 @@ impl PublicKey {
 impl PartialEq for PublicKey {
     fn eq(&self, other: &Self) -> bool {
         // key_id is derived from these fields, so we ignore it.
-        self.typ == other.typ && self.scheme == other.scheme && self.value == other.value
+        self.typ == other.typ
+            && self.scheme == other.scheme
+            && self.keyid_hash_algorithms == other.keyid_hash_algorithms
+            && self.value == other.value
```
erickt, line 691: github isn't letting me comment on lines outside of this patch, but you need to also include `keyid_hash_algorithms` in the `impl hash::Hash for PublicKey` as well.

src/crypto.rs
```diff
@@ -1003,14 +1107,63 @@ mod test {
...
+        });
+
+        let encoded: PublicKey = serde_json::from_value(original.clone()).unwrap();
+        let decoded = serde_json::to_value(&encoded).unwrap();
+
+        assert_eq!(original, decoded);
+    }
+
     #[test]
     fn serde_signature() {
```
erickt, line 1171: This function is getting its keyid changed because it includes keyid_hash_algorithms. Can you add another version that doesn't have keyid_hash_algorithms and confirm the keyid doesn't change?

src/metadata.rs
```diff
@@ -2148,6 +2148,100 @@ mod test {
...
 
+    fn jsn_root_metadata_without_keyid_hash_algos() -> serde_json::Value {
+        json!({
+            "_type": "root",
+            "spec_version": "1.0",
+            "version": 1,
+            "expires": "2017-01-01T00:00:00Z",
+            "consistent_snapshot": false,
+            "keys": {
+                "09557ed63f91b5b95917d46f66c63ea79bdaef1b008ba823808bca849f1d18a1": {
```
erickt: I think this keyid is wrong. If you look above in `serde_root_metadata`, this keyid matches the one that includes `keyid_hash_algorithms`.

kellerb: Updated. Should we verify the key ids when we deserialize?

erickt (approved):

PR #222, opened 2019-11-13
add support for versioned trusted root keys

alpearce:
rust-tuf currently assumes that the trusted_root_keys given when initializing the client are associated with version 1 of the root metadata. This patch changes Client::root_with_pinned to take a version input for the root metadata and start walking the root metadata chain from that possibly later version.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -41,6 +41,7 @@
 //!     Config::default(),
 //!     local,
 //!     remote,
+//!     1,
 //! ).await?;
 //!
 //! let _ = client.update().await?;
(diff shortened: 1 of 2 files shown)


heartsucker (approved):

heartsucker:
Thanks for this!

trishankatdatadog:
We discussed how a TUF client should update the root metadata: roughly speaking, going backward (as in this PR), or going forward (as discussed [here](https://github.com/theupdateframework/specification/issues/56) with @erickt)?

We really should clarify what to do about this in the specification...

Cc @lukpueh @JustinCappos

lukpueh:
Just submitted a PR to resolve inconsistency/ambiguity about root key updates. Feel free to comment in https://github.com/theupdateframework/specification/issues/61.

PR #228, opened 2019-11-15
Setup a github action to build rust-tuf

erickt:
travis seems to have become a little unstable recently. This sets up an experimental github action to build rust-tuf on linux, windows, and OSX.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,0 +1,52 @@
+name: Rust
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+  schedule:
+    - cron: "00 01 * * *"
+
+jobs:
+  ci:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest, macOS-latest]
+        rust:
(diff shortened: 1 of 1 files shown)


heartsucker:
I was considering moving this to CircleCI since I have use them in the past. Is there a reason to use GH actions over CircleCI?

trishankatdatadog:
I have no dog in this fight, but 1st-party >> 3rd-party?

erickt:
@heartsucker: No particular reason, I was curious how it compared to Travis / Appveyor, and I figured it was easy to set up, and I was hoping it'd be more reliable than Travis.

erickt:
@heartsucker: so what do you think? Is this okay, or would you want to use Circle CI? Or would you want to use Github Actions for a period of time until we get Circle CI setup?

heartsucker:
I'm seeing issues with TravisCI for nightly builds right now, so I'm going to merge this and strip out the TravisCI and Appveyor configs to reduce the number of steps in the builds. The one thing we're using is the variations on Windows/GNU and Windows/MSVC. But to be fair, we're not doing anything esoteric, so I'm not worried about it.

PR #231, opened 2019-11-20
TUF 1.0 finds targets metadata by a versioned prefix

### LINKED ISSUE #224: targets metadata should be found by VERSION.FILENAME.EXT
We are currently fetching the targets metadata with HASH.FILENAME.EXT if consistent snapshots are enabled. In section 5.4 of the [TUF Spec], it states that if consistent snapshots is enabled, then the targets metadata is found by fetching VERSION.FILENAME.EXT (e.g. 42.targets.json).

[TUF Spec]: https://github.com/theupdateframework/specification/blob/master/tuf-spec.md#5-detailed-workflows

erickt:
In section 5.4 of the [TUF Spec], it states that if consistent snapshots is enabled, then the targets metadata is found by fetching VERSION.FILENAME.EXT (e.g. 42.targets.json).

Note this modifies `tuf::client::tests::test_version_init` to only generate unversioned metadata if `consistent_snapshot.` is false, and generate versioned snapshot and targets metadata if `consistent_snapshot` is true (which is allowed by the spec).

Closes #224

[TUF Spec]: https://github.com/theupdateframework/specification/blob/master/tuf-spec.md#5-detailed-workflows

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -397,7 +397,7 @@ where
         let (alg, value) = crypto::hash_preference(targets_description.hashes())?;
 
         let version = if self.tuf.root().consistent_snapshot() {
-            MetadataVersion::Hash(value.clone())
+            MetadataVersion::Number(targets_description.version())
         } else {
             MetadataVersion::None
         };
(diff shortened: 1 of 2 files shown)


heartsucker (approved):

heartsucker:
Merging despite spurious Travis failures

PR #233, opened 2019-11-21
Fetch targets by HASH.FILENAME.EXT

### LINKED ISSUE #225: Fetch targets by HASH.FILENAME.EXT
We are currently downloading target files by their filename. However, in section 5.5.2 of the [TUF Spec], it states that if consistent snapshots is enabled, targets are downloaded by fetching HASH.FILENAME.EXT.

[TUF Spec]: https://github.com/theupdateframework/specification/blob/master/tuf-spec.md#5-detailed-workflows

erickt:
In section 5.5.2 of the [TUF Spec], it states that if consistent snapshots is enabled, targets are downloaded by fetching HASH.FILENAME.EXT.

Note that this is layered on top of https://github.com/heartsucker/rust-tuf/pull/231, so I recommend reviewing that first, then reviewing the second commit. 
    
Closes #225
    
[TUF Spec]: https://github.com/theupdateframework/specification/blob/master/tuf-spec.md#5-detailed-workflows

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -469,7 +469,16 @@ where
         target: &'a TargetPath,
     ) -> Result<Box<dyn AsyncRead + Send + Unpin>> {
         let target_description = self.fetch_target_description(target).await?;
-        self.remote.fetch_target(target, &target_description).await
+
+        // According to TUF section 5.5.2, when consistent snapshot is enabled, target files should
+        // be found at `$HASH.FILENAME.EXT`. Otherwise it is stored at `FILENAME.EXT`.
+        if self.tuf.root().consistent_snapshot() {
+            let (_, value) = crypto::hash_preference(target_description.hashes())?;
+            let target = target.with_hash_prefix(value)?;
+            self.remote.fetch_target(&target, &target_description).await
+        } else {
+            self.remote.fetch_target(target, &target_description).await
+        }
     }
 
     async fn lookup_target_description<'a>(
--- a/tests/simple_example.rs
+++ b/tests/simple_example.rs
@@ -1,10 +1,10 @@
 use futures_executor::block_on;
 use tuf::client::{Client, Config, PathTranslator};
-use tuf::crypto::{HashAlgorithm, KeyId, PrivateKey, SignatureScheme};
+use tuf::crypto::{self, HashAlgorithm, KeyId, PrivateKey, SignatureScheme};
 use tuf::interchange::Json;
 use tuf::metadata::{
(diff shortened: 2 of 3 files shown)


heartsucker:
Rebase to make review easier?

erickt:
Rebased! Thanks for merging the other CLs.

tests/simple_example.rs
```diff
@@ -180,3 +180,36 @@ where
...
+    let file_name = match components.pop() {
+        Some(file_name) => file_name,
+        None => {
+            return Err(Error::IllegalArgument("Path cannot be empty".into()));
+        }
+    };
+    components.push(format!("{}.{}", value, file_name));
+
+    TargetPath::new(components.join("/"))
+}
```
ComputerDruid: Rather than do all this, could you just hardcode the path, since it's only used once? That'd probably make the test easier to understand, too.

src/client.rs
```diff
@@ -469,7 +469,25 @@ where
...
+            let mut components = target.components();
+            let file_name = match components.pop() {
+                Some(file_name) => file_name,
+                None => {
+                    return Err(Error::IllegalArgument("Path cannot be empty".into()));
+                }
+            };
+            components.push(format!("{}.{}", value, file_name));
+
+            let target = TargetPath::new(components.join("/"))?;
```
ComputerDruid: It doesn't really feel like this is the right place to be mangling the TargetPath. It seems like this translation is supposed to happen when you translate a VirtualTargetPath into a TargetPath? Does that not work for some reason?

erickt: Note, I factored this out a wee bit to make it a little more readable. Does it read better?

Regarding `VirtualTargetPath`, this is an area where rust-tuf is being a little inconsistent. The metadata is constructed with `VirtualTargetPath`, but `Repository` uses `TargetPath`. I actually think this is the wrong way to go about, the notion of mapping a target path to a real path should really only be done in `FileSystemRepository`, but clients really should just be using `/` separators everywhere (see https://github.com/heartsucker/rust-tuf/issues/236 and https://github.com/theupdateframework/specification/issues/63). I'd like to clean this up if / when we implement #236.

tests/simple_example.rs
erickt: I cleaned this up, like it better?

alpearce:
I took a review pass in ComputerDruid's absence; LGTM.

ComputerDruid (approved):

Issue #235, opened 2019-11-21
Disable appveyor integration


erickt:
It looks like appveyor is still trying to run tests and failing. @heartsucker: Could you disable the integration in https://github.com/heartsucker/rust-tuf/settings/installations? Hopefully that will get our PR builds green.

heartsucker:
This should be done now.

heartsucker:
Config removed again in f4bbb26

Issue #232, opened 2019-11-21
Need a high level interface for creating and changing TUF metadata


erickt:
We need to do a lot of manual work to creating TUF metadata, and generating new versions if we rotate keys or add/remove targets. In comparison, go-tuf has a [repo](https://fuchsia.googlesource.com/third_party/go-tuf/+/refs/heads/master/repo.go) library that has a bunch of simple helper functions. For example, see this [script](https://fuchsia.googlesource.com/third_party/go-tuf/+/refs/heads/master/client/testdata/go-tuf-transition-M4/generate.go) and this [test](https://fuchsia.googlesource.com/third_party/go-tuf/+/refs/heads/master/repo_test.go#567).

cavokz:
Questions:
1. should this be a library+tests?
2. should it also take care of creating keys, if needed (ie using sequoia)?

I see this possible repository creation flow:
1. create TUF repository:
   - root key is existing or new (to be created contextually)
   - at start, all the roles are using the same root key. Maximum simplicity, minimum security.
2. add delegation
   - one delegating key (eg _root_) needs to be accessible for signing the delegation (eg _timestamp_)
   - delegation specific parameters (eg target base path, if creating a _target_ delegation) need to be provided
3. sign existing delegation
   - add a signature to an existing delegation so that delegations requiring multiple signatures can reach the required validity quorum
4. publish

Key rotation would happen in this way:
1. new delegation is added
2. signatures are accumulated until validity quorum is reached
3. signatures are progressively removed from the delegation to be revoked, removal of the last signature removes also the delegation

Need to double check the TUF spec, I don't remember what specifies about rotation. Surely it's not healthy to revoke a delegation without reaching the same validity quorum that qualified it.

erickt:
Hello @cavokz! I'd appreciate any help. I've started sketching out some ideas, but haven't gotten too far yet. Here's roughly the API I think we could use:

```rust
let staging = FileSystemRepository::new("...");
let committed = FileSystemRepository::new("...");

let private_keys = load_keys(...);

let config = Config::default();

// I'd love to call this Repository, but we are using that term already. Maybe we should rename it
// to `Transport` or `Store`?
//
// Anyway, this `Creator` holds all the private keys. It might internally use `Client` to download
// the metadata, and we will use the `local` store as the location for the committed metadata, and
// `staging` as the in progress metadata.
let mut creator = Creator::new(staging, committed);

[...]

PR #237, opened 2019-11-22
Add script to generate metadata for use in compatibility testing. 

alpearce:
(Also check in generated metadata.)

We want to be explicit about making changes to rust-tuf that change the metadata it generates. go-tuf accomplishes this by containing checked-in keys and metadata generated by go-tuf and, in a unit test, asserts that it produces byte-for-byte equivalent metadata. That way, changes to the library that change generated metadata must include corresponding changes to the checked-in metadata to indicate the change in behavior. 

This patch includes the metadata generator and the first set of generated metadata. It uses the same keys as the go-tuf generator so that this metadata can also be used for comparison across implementations. 

Feel free to go wild on the review; this is the biggest hunk of Rust I've written and I'd appreciate a lot of comments.


tests/metadata/generate.rs
```diff
@@ -0,0 +1,262 @@
...
+use tuf::crypto::{HashAlgorithm, KeyType, PrivateKey, SignatureScheme};
+use tuf::interchange::Json;
+use tuf::metadata::{
+    MetadataPath, MetadataVersion, Role, RootMetadataBuilder, SnapshotMetadataBuilder, TargetPath,
+    TargetsMetadataBuilder, TimestampMetadataBuilder, VirtualTargetPath,
+};
+use tuf::repository::{FileSystemRepository, FileSystemRepositoryBuilder, Repository};
+
+// TODO parametrize this better/find it within the build?
+const KEYS_PATH: &str = "./keys.json";
```
erickt, line 17: This should be okay for now. I've been thinking about re-laying out the directory structure, so we could handle improving discovery of the keys.json file then.

tests/metadata/generate.rs
```diff
@@ -0,0 +1,262 @@
...
+    snapshot: Vec<Vec<TestKeyPair>>,
+    timestamp: Vec<Vec<TestKeyPair>>,
+}
+
+fn init_json_keys(path: &str) -> TestKeys {
+    let mut f = File::open(path).expect("failed to open keys file");
+    let mut contents = String::new();
+    f.read_to_string(&mut contents)
+        .expect("failed to read keys");
+    let keys: TestKeys = serde_json::from_str(&contents).expect("serde failed");
```
erickt: You don't need to read this into a string. This should work:

```
let keys: TestKeys = serde_json::from_reader(f).expect("serde failed");
```

tests/metadata/generate.rs
```diff
@@ -0,0 +1,262 @@
...
+    keys.insert("timestamp", json_keys.timestamp[0][0].to_private_key());
+    keys
+}
+
+fn copy_repo(dir: &str, step: u8) {
+    let src = Path::new(dir)
+        .join((step - 1).to_string())
+        .join("repository");
+    let dst = Path::new(dir).join(step.to_string());
+    Command::new("cp")
```
erickt: Could you use `/bin/cp`, just to make sure we don't pick up any old cp executable in the current directory? It should be okay if this only is runnable on a posix-y system, since this is a test script.

tests/metadata/generate.rs
```diff
@@ -0,0 +1,262 @@
...
+fn init_role_keys(json_keys: &TestKeys) -> RoleKeys {
+    let mut keys = HashMap::new();
+    keys.insert("root", json_keys.root[0][0].to_private_key());
+    keys.insert("snapshot", json_keys.snapshot[0][0].to_private_key());
+    keys.insert("targets", json_keys.targets[0][0].to_private_key());
+    keys.insert("timestamp", json_keys.timestamp[0][0].to_private_key());
+    keys
+}
+
+fn copy_repo(dir: &str, step: u8) {
```
erickt, line 71: Generally speaking I'd prefer this call some rust function to copy the directories, since we might not guarantee that `cp` is available. Eventually I'd like to turn this into a test library, so that we can add a test that will complain if we accidentally change the metadata format without regenerating the test metadata, like we're doing in [go-tuf](https://fuchsia.googlesource.com/third_party/go-tuf/+/refs/heads/master/client/interop_test.go#32).

Unfortunately, I'm guessing you discovered there's no standard library way for copying directories. We might be able to pull in a crate for this ([fs_extra](https://crates.io/crates/fs_extra) seems popular, but we need to check out it's quality), but for now I say we punt for now and solve this later for when we create this identity test.

So in short, could you add a `// TODO replace this with pure rust library so it's portable across OSes`?

tests/metadata/generate.rs
```diff
@@ -0,0 +1,262 @@
...
+    repo.store_metadata(&root_path, &MetadataVersion::Number(version), &root)
+        .await
+        .unwrap();
+    repo.store_metadata(&root_path, &MetadataVersion::None, &root)
+        .await
+        .unwrap();
+}
+
+// adds a target and updates the non-root metadata files.
+async fn add_target(
```
erickt, line 121: More evidence we need a [helper library for generating metadata](https://github.com/heartsucker/rust-tuf/issues/232).

tests/metadata/generate.rs
```diff
@@ -0,0 +1,262 @@
...
+
+    // Timestamp doesn't require a version even in consistent_snapshot.
+    repo.store_metadata(&timestamp_path, &MetadataVersion::None, &timestamp)
+        .await
+        .unwrap();
+}
+
+async fn generate_repos(dir: &str, consistent_snapshot: bool) -> tuf::Result<()> {
+    // Create initial repo.
+    // TODO: @erickt, this impl is very tied to the json keys file. Do we need it to be more flexible?
```
erickt: This should be fine for now.

tests/metadata/generate.rs
```diff
@@ -0,0 +1,262 @@
...
+        // Initialize new repo and copy the files from the previous step.
+        let dir_i = Path::new(dir).join(i.to_string());
+        let repo = FileSystemRepositoryBuilder::new(dir_i)
+            .metadata_prefix(Path::new("repository"))
+            .targets_prefix(Path::new("repository").join("targets"))
+            .build()
+            .unwrap();
+        copy_repo(dir, i);
+
+        let root_signer: Option<PrivateKey> = match r {
```
erickt: Nit: I think you can leave off `Option<PrivateKey>`.

erickt (changes requested):
A few things other things:

* First, it looks like you hit a [test failure](https://github.com/heartsucker/rust-tuf/pull/237/checks?check_run_id=315130477). You can ignore the appveyor, that's [going away](https://github.com/heartsucker/rust-tuf/issues/235).
* Darn, I thought I added a mode FileSystemBuilder that lets us pretty print the metadata. That makes it much easier to review. I filed https://github.com/heartsucker/rust-tuf/issues/238 to track implementing this.

tests/metadata/generate.rs
alpearce: Ack.

tests/metadata/generate.rs
alpearce: Done.

tests/metadata/generate.rs
alpearce: Done.

tests/metadata/generate.rs
alpearce: Added TODO.

tests/metadata/generate.rs
alpearce: Ack.

tests/metadata/generate.rs
alpearce: Done.

alpearce:
Updated:
  - @erickt's comments
  - use a constant expiration so metadata will be static
  - regenerate metadata with the above

heartsucker:
Nice! We should fix #238 before merging this one otherwise reviews are going to be impossible.

alpearce:
Borrowed the "linkify-metadata" script from go-tuf and regenerated symlinked metadata.

tests/metadata/generate.rs
cavokz: @alpearce is it ok for you if I attempt at the library part or is it something you would like to do?

erickt:
A couple more things:

* I uploaded a pretty printing interchange in https://github.com/heartsucker/rust-tuf/pull/241, could you rebase your patch on top of this one and use it to regenerate the metadata? While you do that, can you squash all the patches together, then factor out the metadata into it's own patch? That should make it easier to review the metadata generation script from the actual metadata.
* I used your PR to test out that `JsonPretty` worked (thanks!) I was able to successfully parse step `0`, but I ran into a bug in step `1`. It is missing the `0` target. Instead, it should look like [this](https://github.com/heartsucker/rust-tuf/blob/develop/tests/interop/fuchsia-go-tuf-5527fe/consistent-snapshot-false/1/repository/targets.json).
* You don't need to do this now, but it'd be nice if we could migrate this script into a separate independent cargo toml. I'd be fine if you want us to do this in a future PR though to get this landed.

tests/metadata/generate.rs
alpearce: Fine with me!

alpearce:
Prettified, squashed, separated metadata, fixed bug, added TODO in Cargo.toml.

erickt (approved):

PR #241, opened 2019-11-27
Add a json pretty printing interchange

### LINKED ISSUE #238: Extend FileSystemRepositoryBuilder to support pretty printing metadata
In https://github.com/heartsucker/rust-tuf/pull/237, @alpearce is writing a generator for the interop test suite, but unfortunately FileSystemRepository only writes out minified json. It would be more helpful for reviews if we could also generate pretty printed metadata, so it can be easily reviewed by humans.

erickt:
This adds an interchange called `JsonPretty`, which makes it easier for humans to review metadata changes. It is identical to `tuf::interchange::Json`, except for the implementation of `to_writer`. Instead of writing out the canonical form, it writes out a pretty printed form of the data.

In addition, it changes `SnapshotMetadataBuilder::insert_metadata_with_path` and `TimestampMetadataBuilder::from_snapshot` to use `D::to_writer`. This makes sure that the hash of these metadata files are computed from the format written to disk, instead of the canonical format.

Closes: #238

### DIFF
--- a/src/interchange/cjson/mod.rs
+++ b/src/interchange/cjson/mod.rs
@@ -264,7 +264,7 @@ impl DataInterchange for Json {
     /// let arr = vec![1, 2, 3];
     /// let mut buf = Vec::new();
     /// Json::to_writer(&mut buf, &arr).unwrap();
-    /// assert!(&buf == b"[1, 2, 3]" || &buf == b"[1,2,3]");
+    /// assert_eq!(&buf, b"[1,2,3]");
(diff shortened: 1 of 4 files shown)


heartsucker (commented):
Does this also pretty print structs? The docstring tests don't show that, and that might be helpful.

erickt:
Good idea. I extended the tests to include checks for all the json variants.

alpearce:
Looks good to me.

PR #242, opened 2019-11-28
Add new Client constructors, deprecate old ones

### LINKED ISSUE #229: Client::with_root_pinned should take public keys, not key ids
As @ComputerDruid [pointed out](https://github.com/heartsucker/rust-tuf/pull/213#discussion_r346988780), Client::with_root_pinned should be taking the public keys used to sign the metadata, rather than trusting the initial root metadata has the correct mapping from key id to key. Otherwise, if we removed recomputing the key ids (see https://github.com/theupdateframework/tuf/issues/848#issuecomment-551315734), then a malicious server could swap out the public keys but preserve the key ids and we wouldn't notice.

erickt:
This patch changes how Clients are created. It adds the following
functions:

* `Client::from_local` - use the specified root version from the local repository as our initial trusted root.
* `Client::from_pinned_root_keyids` - use the specified root version, threshold, and keyids to trust a root fetched from the local or remote repository.
* `Client::from_pinned_root_keys` - use the specified root version, threshold, and public keys to trust a root fetched from the local or remote repository.
* `Client::from_pinned_root` - use the specified root metadata as the initial trusted root.

This deprecates the old constructors:

* `Client::new`
* `Client::with_root_pinned`

Closes: #229

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -36,7 +36,7 @@
 //! .user_agent("rustup/1.4.0")
 //! .build();
 //!
-//! let mut client = Client::with_pinned_root_keyids(
+//! let mut client = Client::with_trusted_root_keyids(
 //!     Config::default(),
 //!     &MetadataVersion::Number(1),
 //!     1,
@@ -142,17 +142,17 @@ where
     /// **WARNING**: This method offers weaker security guarantees than the related method
     /// `with_root_pinned`.
     ///
-    /// **DEPRECATED**: This has been replaced with [Client::from_local].
-    #[deprecated(note = "use Client::from_local method instead")]
+    /// **DEPRECATED**: This has been replaced with [`Client::with_trusted_local`].
+    #[deprecated(note = "use Client::with_trusted_local method instead")]
     pub async fn new(config: Config<T>, local: L, remote: R) -> Result<Self> {
-        Client::with_trusted_local(config, &MetadataVersion::Number(1), local, remote).await
+        Client::with_trusted_local(config, local, remote).await
     }
 
     /// Create a new TUF client. It will attempt to load initial root metadata the local and remote
     /// repositories using the provided key IDs to pin the verification.
     ///
-    /// **DEPRECATED**: This has been replaced with [Client::with_pinned_root_keyids].
-    #[deprecated(note = "use Client::with_pinned_root_keyids method instead")]
+    /// **DEPRECATED**: This has been replaced with [`Client::with_trusted_root_keyids`].
+    #[deprecated(note = "use Client::with_trusted_root_keyids method instead")]
     pub async fn with_root_pinned(
         trusted_root_keys: &[KeyId],
         config: Config<T>,
@@ -160,7 +160,7 @@ where
         remote: R,
         version: u32,
     ) -> Result<Self> {
-        Client::with_pinned_root_keyids(
+        Client::with_trusted_root_keyids(
             config,
             &MetadataVersion::Number(version),
             1,
@@ -175,8 +175,9 @@ where
     /// repo and use it as the initial trusted root metadata, or it will return an error if it
     /// cannot do so.
     ///
-    /// **WARNING**: This method offers weaker security guarantees than the related method
-    /// `with_pinned_root_keys`.
+    /// **WARNING**: This is trust-on-first-use (TOFU) and offers weaker security guarantees than
+    /// the related methods [`Client::with_trusted_root`], [`Client::with_trusted_root_keys`] and
+    /// [`Client::with_trusted_root_keyids`].
     ///
     /// # Examples
     ///
@@ -218,7 +219,6 @@ where
     ///
     /// let client = Client::with_trusted_local(
     ///     Config::default(),
-    ///     &root_version,
     ///     local,
     ///     remote,
     /// ).await?;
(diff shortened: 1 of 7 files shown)


ComputerDruid (commented):
FYI, from `cargo rustdoc`:

```
warning: `[Client::from_root_pinned_keyids]` cannot be resolved, ignoring it...
   --> src/client.rs:154:54
    |
154 |     /// **DEPRECATED**: This has been replaced with [Client::from_root_pinned_keyids].
    |                                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot be resolved, ignoring
    |
    = note: `#[warn(intra_doc_link_resolution_failure)]` on by default
    = help: to escape `[` and `]` characters, just add '\' before them like `\[` or `\]`

    Finished dev [unoptimized + debuginfo] target(s) in 0.11s
```

I can't seem to comment on the right line because the diff isn't loading, not sure what's up with that.

erickt:
Thanks, I just pushed up a fix for the deprecation notice. It also looks like the diff view is working again.

src/client.rs
```diff
@@ -139,12 +141,101 @@ where
...
     pub async fn new(config: Config<T>, local: L, remote: R) -> Result<Self> {
+        Client::from_local(config, &MetadataVersion::Number(1), local, remote).await
+    }
+
+    /// Create a new TUF client. It will attempt to load initial root metadata the local and remote
+    /// repositories using the provided key IDs to pin the verification.
+    ///
+    /// **DEPRECATED**: This has been replaced with [Client::from_pinned_root_keyids].
+    #[deprecated(note = "use Client::from_pinned_root_keyids method instead")]
+    pub async fn with_root_keyids_pinned(
```
ComputerDruid: If you're keeping this deprecated to not break callers, you should leave the name as with_root_pinned

src/client.rs
```diff
@@ -160,38 +251,272 @@ where
...
+        let root_version = MetadataVersion::Number(trusted_root.version());
+
+        local
+            .store_metadata(&root_path, &root_version, &trusted_root)
+            .await?;
+
+        // Also store this root metadata as the latest version.
+        local
+            .store_metadata(&root_path, &MetadataVersion::None, &trusted_root)
+            .await?;
```
ComputerDruid: Same here; I think this will break offline root version walking.

src/client.rs
```diff
@@ -160,38 +251,272 @@ where
...
+            let root_version = MetadataVersion::Number(root.version());
+            local
+                .store_metadata(&root_path, &root_version, &root)
+                .await?;
+
+            // Also store this root metadata as the latest version.
+            local
+                .store_metadata(&root_path, &MetadataVersion::None, &root)
+                .await?;
+        }
```
ComputerDruid: I think this should be conditional on whether you're increasing the version somehow. Can you add tests for this case to each of these constructors to show that these don't roll back root?

src/client.rs
```diff
@@ -160,38 +251,272 @@ where
...
+    ///     root_threshold,
+    ///     once(&public_key),
+    ///     local,
+    ///     remote,
+    /// ).await?;
+    /// # Ok(())
+    /// # })
+    /// # }
+    /// ```
+    pub async fn from_pinned_root_keys<'a, I>(
```
ComputerDruid: I think "from_pinned_root_keys" isn't a very descriptive name for this constructor, since it doesn't really come "from" the root keys. It's really `from_local_or_remote_with_pinned_root_keys`, but that name is probably too long. `with_pinned_root_keys` seems like an acceptable shorthand to me, but my feelings aren't that strong on the matter. Feel free to get a second opinion.

src/tuf.rs
```diff
@@ -26,27 +26,47 @@ pub struct Tuf<D: DataInterchange> {
 }
 
 impl<D: DataInterchange> Tuf<D> {
-    /// Create a new `TUF` struct from a known set of pinned root keys that are used to verify the
-    /// signed metadata.
-    pub fn from_root_pinned<'a, I>(
-        mut signed_root: SignedMetadata<D, RootMetadata>,
+    /// Create a new `TUF` struct from a known set of pinned root key ids that are used to verify
+    /// the signed metadata.
+    pub fn from_root_keyids_pinned<'a, I>(
```
ComputerDruid: Maybe this should be named `from_root_with_pinned_keyids`?

src/client.rs
erickt: I think it was a mistake to fold this change into this patch. I'll remove it.

src/tuf.rs
erickt: I like it, I'll switch to it.

src/client.rs
erickt: `with_pinned_root_keys` seems fine to me. I also updated `Tuf` to reflect these names.

src/client.rs
```diff
@@ -418,36 +419,35 @@ where
...
-        Ok(Client {
-            tuf,
-            config,
-            local,
-            remote,
-        })
+        Ok(client)
     }
 
     /// Create a new TUF client. It will use trust this initial root metadata.
```
ComputerDruid: s/use trust/trust/

src/client.rs
```diff
@@ -418,36 +419,35 @@ where
...
+            client
                 .store_metadata(&root_path, &root_version, &root)
-                .await?;
+                .await;
 
-            // Also store this root metadata as the latest version.
-            local
-                .store_metadata(&root_path, &MetadataVersion::None, &root)
-                .await?;
+            // FIXME: should we also the root as `MetadataVersion::None`?
```
ComputerDruid: s/also/also store/

src/client.rs
```diff
@@ -498,25 +498,13 @@ where
     /// # })
     /// # }
     /// ```
-    pub async fn from_pinned_root(
+    pub async fn with_pinned_root(
```
ComputerDruid: maybe "trusted" is a better name than "pinned"

src/tuf.rs
```diff
@@ -46,12 +46,12 @@ impl<D: DataInterchange> Tuf<D> {
...
         }
 
-        Self::from_root(signed_root)
+        Self::with_root(signed_root)
     }
 
     /// Create a new `TUF` struct from a known set of pinned root keys that are used to verify the
     /// signed metadata.
-    pub fn from_root_keys_pinned<'a, I>(
+    pub fn with_root_with_pinned_keys<'a, I>(
```
ComputerDruid: Can we call this `from_root_with_trusted_root_keys`?

src/tuf.rs
```diff
@@ -60,14 +60,14 @@ impl<D: DataInterchange> Tuf<D> {
...
+        Self::with_root(signed_root)
     }
 
     /// Create a new `TUF` struct from a piece of metadata that is assumed to be trusted.
     ///
     /// **WARNING**: This is trust-on-first-use (TOFU) and offers weaker security guarantees than
-    /// the related method `from_root_keys_pinned` and `from_root_keyids_pinned`.
-    pub fn from_root(signed_root: SignedMetadata<D, RootMetadata>) -> Result<Self> {
+    /// the related method `with_root_with_pinned_keys` and `with_root_with_pinned_keyids`.
+    pub fn with_root(signed_root: SignedMetadata<D, RootMetadata>) -> Result<Self> {
```
ComputerDruid: `from_trusted_root` to imply where the trust is starting?

src/tuf.rs
```diff
@@ -28,7 +28,7 @@ pub struct Tuf<D: DataInterchange> {
 impl<D: DataInterchange> Tuf<D> {
     /// Create a new `TUF` struct from a known set of pinned root key ids that are used to verify
     /// the signed metadata.
-    pub fn from_root_keyids_pinned<'a, I>(
+    pub fn with_root_with_pinned_keyids<'a, I>(
```
ComputerDruid: You can have this delegate to `from_root_with_trusted_root_keys` and then inline this into client.

src/client.rs
```diff
@@ -145,22 +145,22 @@ where
     /// **DEPRECATED**: This has been replaced with [Client::from_local].
```
ComputerDruid: please fix

ComputerDruid (approved):
Overall looks good but I think you could improve the naming of these constructors.

src/tuf.rs
erickt: I lifted this logic up into `Client`.

PR #243, opened 2019-11-28
Fix documentation of SignedMetadata::add_signature

cavokz:


heartsucker (approved):
Thanks for the fix :)

PR #244, opened 2019-11-28
Make git ignore .* files

cavokz:
The reason is that vim (my text editor) uses files starting with '.' for autosave and recovery, these files are hidden by default by 'ls' and most other Linux applications.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 target
 Cargo.lock
 *.rs.bk
+.*


heartsucker:
If you're using `vim`, you probably want this to ignore `*.swp` and `*.swo` files. Using `.*` is too broad. You can also set this up globally:

Add this to your `~/.gitconfig`:
```ini
[core]
    excludesfile = ~/.gitignore
```

And then this to `~/.gitignore`
```
*.swp
*.swo
```

cavokz:
Good point, thanks!

PR #246, opened 2019-12-03
Identity test with checked in metadata

alpearce:
This patch adds a step to the interop_test that makes sure we can read in the generated metadata. At some point, the interop test will do the work of comparing the metadata exactly so that we can catch changes to the way we generate metadata.

Bugs found and fixed:
- copy repos function should use output() not spawn() to make sure the call completes
- need to explicitly set the metadata versions for each step
- when rotating the root key, both the old and new roots need to sign the metadata

Tested: fuchsia_rust_tuf_identity_consistent_snapshot_false passes

The regenerated metadata is a separate patch to make review easier.

### DIFF
--- a/tests/interop/main.rs
+++ b/tests/interop/main.rs
@@ -90,7 +90,7 @@ fn fuchsia_go_tuf_transition_m4_consistent_snapshot_true() {
 
 // Tests to catch changes to the way we generate metadata.
 #[test]
-fn fuchsia_rust_tuf_identity_consistent_snapshot_false() {
+fn rust_tuf_identity_consistent_snapshot_false() {
     test_key_rotation(
         Path::new("tests")
             .join("metadata")
@@ -100,7 +100,7 @@ fn fuchsia_rust_tuf_identity_consistent_snapshot_false() {
 
 #[test]
 #[ignore] // FIXME: Blocked on #225
-fn fuchsia_rust_tuf_identity_consistent_snapshot_true() {
+fn rust_tuf_identity_consistent_snapshot_true() {
     test_key_rotation(
         Path::new("tests")
             .join("metadata")
--- a/tests/interop/main.rs
+++ b/tests/interop/main.rs
@@ -85,7 +85,27 @@ fn fuchsia_go_tuf_transition_m4_consistent_snapshot_true() {
             .join("interop")
             .join("fuchsia-go-tuf-transition-M4")
             .join("consistent-snapshot-true"),
-    )
+    );
+}
+
+// Tests to catch changes to the way we generate metadata.
+#[test]
+fn fuchsia_rust_tuf_identity_consistent_snapshot_false() {
+    test_key_rotation(
+        Path::new("tests")
+            .join("metadata")
+            .join("consistent-snapshot-false"),
+    );
+}
+
+#[test]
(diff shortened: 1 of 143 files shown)


tests/interop/main.rs
```diff
@@ -87,7 +87,27 @@ fn fuchsia_go_tuf_transition_m4_consistent_snapshot_true() {
             .join("interop")
             .join("fuchsia-go-tuf-transition-M4")
             .join("consistent-snapshot-true"),
-    )
+    );
+}
+
+// Tests to catch changes to the way we generate metadata.
+#[test]
+fn fuchsia_rust_tuf_identity_consistent_snapshot_false() {
```
ComputerDruid: As far as I can tell; these are for rust-tuf in general, so no need for the word "fuchsia" here. (for the go-tuf metadata, the metadata is from the fuchsia fork of go-tuf, hence the name)

ComputerDruid (commented):
Looks good overall, but it looks like one of the new tests failed in CI.

erickt:
You got hit by git on windows replacing `\n` with `\r\n`, which messes with TUF expecting the metadata file sizes to be stable. You could copy `tests/interop/fuchsia-go-tuf-transition-M4/.gitattributes` into your metadata directory. In the meantime, I'll see if I can lift up this .gitattributes files to the top level to make it apply everywhere.

trishankatdatadog:
Hah, yeah, we've seen this before! We [normalize line endings](https://github.com/DataDog/integrations-core/blob/137761f2d4540ab9f3c0e7a10e1658ff4f74c882/datadog_checks_dev/datadog_checks/dev/tooling/signing.py#L83) when calling in-toto for this reason...

tests/interop/main.rs
alpearce: Removed the "fuchsia"

tests/interop/main.rs
```diff
@@ -85,7 +85,27 @@ fn fuchsia_go_tuf_transition_m4_consistent_snapshot_true() {
...
+fn rust_tuf_identity_consistent_snapshot_false() {
+    test_key_rotation(
+        Path::new("tests")
+            .join("metadata")
+            .join("consistent-snapshot-false"),
+    );
+}
+
+#[test]
+#[ignore] // FIXME: Blocked on #225
```
erickt, line 102: #225 has landed, so you shouldn't need to ignore this test.

tests/metadata/generate.rs
```diff
@@ -77,7 +77,7 @@ fn copy_repo(dir: &str, step: u8) {
         .arg("-r")
         .arg(src.to_str().unwrap())
         .arg(dst.to_str().unwrap())
-        .spawn()
+        .output()
```
erickt, line 80: Was this switched for debugging purposes? I think we could switch back to spawn if you aren't doing anything with the stdout.

ComputerDruid (approved):

erickt (commented):
Other than that, LGTM

tests/metadata/generate.rs
alpearce: This was intentional - spawn didn't wait for the command to complete, so there was a race condition.

PR #247, opened 2019-12-03
turn off git normalizing line endings

erickt:


.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Never modify lines endings
```
ComputerDruid: Can you add a note to this comment that we're disabling it because line normalization can break integration tests if `core.autocrlf=true`?

ComputerDruid (approved):

PR #251, opened 2019-12-05
Fix directory existence check

cavokz:
The _if [[ ... ]]_ construct is a bash-ism and fails to work if /bin/sh does not point to bash. An alternative fix is to use /bin/bash as shebang instead of /bin/sh. Feel free to edit as you prefer.


erickt:
thanks!

PR #255, opened 2019-12-10
Update docs for keys generation.

cavokz:
'cargo install tuf' does not work (probably outdated).

Running 'cargo run tuf' instead triggers tests/metadata/generate.rs
which panics for not being able to load keys. Again, misleading at best.


erickt:
Thanks again!

PR #259, opened 2019-12-13
Enable IPv6 zone ID support for HttpRepository

wittrock:
Accomplished by switching to `http::Uri` rather than `url::Url`. This
is somewhat unfortunate, as the ergonomics of `http::Uri` are notably
worse, but the result is that zone IDs work.

For API compatibility purposes, the changes manifest as a new struct
called HttpRepository2 and HttpRepositoryBuilder2. In a subsequent
commit we can delete HttpRepository and rename HttpRepository2.

### DIFF
--- a/src/repository.rs
+++ b/src/repository.rs
@@ -511,12 +511,12 @@ fn extend_uri(uri: Uri, prefix: &Option<Vec<String>>, components: &[String]) -> 
         .split("/")
         .map(String::from)
         .collect::<Vec<_>>();
-    let mut new_path_elements: Vec<String> = vec![];
+    let mut new_path_elements: Vec<&str> = vec![];
 
     if let Some(ref prefix) = prefix {
-        new_path_elements.extend(prefix.iter().cloned());
+        new_path_elements.extend(prefix.iter().map(String::as_str));
     }
-    new_path_elements.extend_from_slice(components);
+    new_path_elements.extend(components.iter().map(String::as_str));
 
     // Urlencode new items to match behavior of PathSegmentsMut.extend from
     // https://docs.rs/url/2.1.0/url/struct.PathSegmentsMut.html
--- a/src/repository.rs
+++ b/src/repository.rs
@@ -265,7 +265,7 @@ where
 
             Ok(())
         }
-        .boxed()
+            .boxed()
     }
 
     /// Fetch signed metadata.
@@ -297,7 +297,7 @@ where
 
             Ok(D::from_slice(&buf)?)
         }
-        .boxed()
+            .boxed()
     }
 
     fn store_target<'a, R>(
@@ -322,7 +322,7 @@ where
 
             Ok(())
         }
-        .boxed()
+            .boxed()
     }
 
     fn fetch_target<'a>(
@@ -349,7 +349,7 @@ where
 
             Ok(reader)
         }
-        .boxed()
+            .boxed()
     }
 }
 
@@ -373,7 +373,7 @@ where
     C: Connect + Sync + 'static,
     D: DataInterchange,
 {
-    url: Url,
+    uri: Uri,
     client: Client<C>,
     interchange: PhantomData<D>,
     user_agent: Option<String>,
@@ -390,7 +390,20 @@ where
     /// Create a new repository with the given `Url` and `Client`.
     pub fn new(url: Url, client: Client<C>) -> Self {
         HttpRepositoryBuilder {
-            url: url,
+            uri: url.to_string().parse::<Uri>().unwrap(), // This is dangerous, but will only exist for a short time as we migrate APIs.
+            client: client,
+            interchange: PhantomData,
+            user_agent: None,
+            metadata_prefix: None,
+            targets_prefix: None,
+            min_bytes_per_second: 4096,
+        }
+    }
+
+    /// Create a new repository with the given `Url` and `Client`.
+    pub fn new_with_uri(uri: Uri, client: Client<C>) -> Self {
+        HttpRepositoryBuilder {
+            uri: uri,
             client: client,
             interchange: PhantomData,
             user_agent: None,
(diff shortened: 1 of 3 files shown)


src/repository.rs
```diff
@@ -537,7 +538,318 @@ where
...
+    }
+
+    let mut path_split = modified_path
+        .split("/")
+        .map(String::from)
+        .collect::<Vec<_>>();
+    let mut new_path_elements: Vec<String> = vec![];
+
+    if let Some(ref prefix) = prefix {
+        new_path_elements.extend(prefix.iter().cloned());
```
ComputerDruid: It doesn't seem like you would need `.cloned()` here if new_path_elements were a `Vec<&str>`

src/repository.rs
```diff
@@ -537,7 +538,318 @@ where
...
+    }
+    new_path_elements.append(&mut components.to_vec());
```
ComputerDruid: `to_vec` copies the list, but you should be able to just use the slice directly by using `extend_from_slice` or `extend` instead of `append`.

src/repository.rs
```diff
@@ -537,7 +538,318 @@ where
...
+    D: DataInterchange,
+{
+    uri: Uri,
+    client: Client<C>,
+    user_agent: String,
+    metadata_prefix: Option<Vec<String>>,
+    targets_prefix: Option<Vec<String>>,
+    min_bytes_per_second: u32,
+    interchange: PhantomData<D>,
+}
```
ComputerDruid: As far as I can tell, `HttpRepository.url` only appears in the API in `HttpRepositoryBuilder::new`. So instead of having a whole `HttpRespository2`, couldn't you just have 2 constructors for `HttpRepositoryBuilder`, which internally always converts to `http::Uri`?

src/repository.rs
```diff
@@ -537,7 +538,318 @@ where
...
+    .add(b'/')
+    .add(b':')
+    .add(b';')
+    .add(b'=')
+    .add(b'@')
+    .add(b'[')
+    .add(b'\\')
+    .add(b']')
+    .add(b'^')
+    .add(b'|');
```
ComputerDruid: This looks wrong; did you mean `#?{}`?

src/repository.rs
```diff
@@ -537,7 +538,318 @@ where
...
+        new_path_elements.extend(prefix.iter().cloned());
+    }
+    new_path_elements.append(&mut components.to_vec());
+
+    // Urlencode new items to match behavior of PathSegmentsMut.extend from
+    // https://docs.rs/url/2.1.0/url/struct.PathSegmentsMut.html
+    let encoded_new_path_elements: Vec<String> = new_path_elements
+        .into_iter()
+        .map(|path_segment| utf8_percent_encode(&path_segment, URLENCODE_PATH).collect())
+        .collect();
```
ComputerDruid: you shouldn't need a `collect()` here, since extend accepts an iterator.

src/repository.rs
```diff
@@ -774,6 +1086,131 @@ mod test {
...
+        let prefix = Some(vec![String::from("prefix")]);
+        let components = [String::from("chars_to_encode<>!")];
+        let uri = base_uri.parse::<Uri>().unwrap();
+        let extended_uri = extend_uri(uri, &prefix, &components)
+            .expect("correctly generated a URI with a zone id");
+
+        let url =
+            http_repository_extend_using_url(Url::parse(base_uri).unwrap(), &prefix, &components);
+
+        assert_eq!(url.to_string(), extended_uri.to_string());
```
ComputerDruid, line 899: Can you also assert that these match a fixed string? It makes sense to check that url agrees with you, but it'd also be nice to see what it is that you're expecting while reading the test.

ComputerDruid (changes requested):

src/repository.rs
wittrock: Done.

src/repository.rs
wittrock: `std::iter::Extend<&std::string::String>` is not implemented for `std::vec::Vec<&str>`

Is there a conversion I could do that's more idiomatic?

src/repository.rs
wittrock: Yup, done.

src/repository.rs
wittrock: Done.

src/repository.rs
wittrock: Done.

src/repository.rs
wittrock: Done.

wittrock:
The CI error looks like an infra configuration problem. How are those normally handled with this repo?

src/repository.rs
wittrock: I ended up needing to do `new_path_elements.extend(prefix.iter().map(AsRef::as_ref));`, and I'm not sure the lack of copying is worth the decrease in readability. What do you think?

ComputerDruid: yeah, I think it's worth it. In general, unless you're "storing the string away for later", it's much more idiomatic to work with `&str` in rust. You don't necessarily need to push that all the way back to the caller of this function (although you could if you really like iterators), but at least in the body of this function you can work with `&str` until you're ready to do the final "format everything back to string".

I thought you could just do `.iter().map(|s| &*s)` or `.iter().map(|&s| &s)`, but apparently type inference doesn't work in this position.

What does seem to work is `.iter().map(String::as_str)` which IMO is a fairly clear way to say "turn `String`s into `str`s": https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=bdb27c4baf74e45c295b4de0756c1a3e

ComputerDruid:
> The CI error looks like an infra configuration problem. How are those normally handled with this repo?

Opened https://github.com/heartsucker/rust-tuf/pull/260

src/repository.rs
wittrock: Done.

ComputerDruid (approved):

PR #262, opened 2020-01-02
Drop MetadataVersion::Hash, resolve #254

### LINKED ISSUE #254: Remove MetadataVersion::Hash
According to the TUF-1.0 spec, metadata is only ever fetched unversioned, or by a specific version number, including delegations. Therefore we should be able to remove `MetadataVersion::Hash`.

cavokz:


src/client.rs
```diff
@@ -871,12 +871,6 @@ where
                 Err(e) => return (delegation.terminating(), Err(e)),
             };
 
-            let version = if self.tuf.root().consistent_snapshot() {
-                MetadataVersion::Hash(value.clone())
```
erickt, line 875: I think we want this conditional, just returning `MetadataVersion::Number(role_meta.version())` when consistent snapshot is enabled.

erickt (changes requested):
Thanks for this! Other than my in-line comment, can you also:

* check if python-tuf is fetching delegations by version number, just to be safe?
* add a test to make sure we are fetching delegations by version number. Simplest would be to extend src/client.rs's `test_versioned_init` with a delegation. You can find how to make a delegation by looking at tests/integration.rs.

cavokz:
I addressed the inline comment, I'll handle the test extension in a separate commit. I still need to understand a few more things. Thanks!

erickt (approved):

erickt:
This was incorporated into another PR that has since landed. Thanks!

PR #267, opened 2020-01-24
remove deprecated functions

erickt:
Fuchsia has migrated away from these functions, so we don't need to keep them around anymore.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -136,41 +136,6 @@ where
     R: Repository<D>,
     T: PathTranslator,
 {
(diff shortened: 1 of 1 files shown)


heartsucker:
I had realized you had deprecated `with_root_pinned`. I think that should stay in since the specific use case with that was hard coding key IDs into a binary. I think that should stay in unless you have a specific reason to remove that.

erickt:
We still have the functionality of `with_root_pinned`, the function just got renamed to `Client::with_trusted_root_keyids` to distinguish it from creating clients with `PublicKey`s or with an initial trusted `SignedMetadata<RootMetadata>`.

PR #271, opened 2020-01-29
update to hyper 0.13 and http 0.2

tamird:
@erickt


erickt:
Looks good to me, but we need to hold off on this until we're ready to update Fuchsia.

Issue #119, opened 2017-07-26
Split `Repository` trait into local and remote components


heartsucker:
Because we trust data once its in the local repo, and we don't need to reverify it on every fetch.

erickt:
For repositories like `FileSystemRepository`, I assume we don't need to check for trickle attacks, but do you think we ought to verify the hash on every read?

erickt:
Circling back, yes we should verify the hash on every read in `FileSystemRepository`. On fuchsia, we store our trusted root in a read only location, so we can cryptographically verify that all the metadata and target files are based off of our initial trusted root. Since we want to protect against an attacker being able to manipulate our mutable storage, we need to check the file is as expected when we open it up.

EphemeralRepository however does not need to verify the hash, since if an attacker can modify our TUF process's memory, then they can probably skip verification altogether.

heartsucker:
If we want to always verify from a local repo, we can close this. Since the ephemeral repo is only used for testing, it's ok if it's not particularly efficient (not that hashing is slow, but it's an unnecessary optimization)

Issue #253, opened 2019-12-06
Client::update_root should verify fetched metadata is expected version


erickt:
When `tuf::Client::update_root` downloads root metadata from a server, it's currently not checking that that metadata's version matches the expected version. So this means if we are currently on version 5.root.json, when we try to download 6.root.json the server could actually give us version 7. We should make sure the metadata matches the expected version.

erickt:
Here is the relevant section of the spec:

> 1.4. Check for a rollback attack. The version number of the trusted root metadata file (version N) must be less than or equal to the version number of the new root metadata file (version N+1). Effectively, this means checking that the version number signed in the new root metadata file is indeed N+1. If the version of the new root metadata file is less than the trusted metadata file, discard it, abort the update cycle, and report the rollback attack. On the next update cycle, begin at step 0 and version N of the root metadata file.

PR #274, opened 2020-02-03
Split repository traits, Preserve metadata hashes

wellsie1116:
This series of changes:
 * Splits the Repository trait into a read half (RepositoryProvider) and write half (RepositoryStorage), resolving #119 
 * Moves hash and max length verification out of the repository trait implementations and into a Repository struct that wraps a repository implementation, allowing this crate to enforce those constraints even with externally implemented `RepositoryProvider`s.
 * Preserves the original bytes of metadata long enough to store it locally, allowing hash checks to pass when re-reading locally stored metadata.


src/client.rs
```diff
@@ -191,13 +195,18 @@ where
     /// # })
     /// # }
     /// ```
-    pub async fn with_trusted_local(config: Config<T>, local: L, remote: R) -> Result<Self> {
+    pub async fn with_trusted_local(
+        config: Config<T>,
+        local: impl Into<Repository<L, D>>,
```
erickt: Since this is part of the public interface, can you switch to using where clauses? Using `impl Into` prevents users from turbofish-ing this type, and since this is a fairly generic type, it's possible we might need a turbofish for some complex situations.

src/client.rs
```diff
@@ -191,13 +195,18 @@ where
...
+        remote: impl Into<Repository<R, D>>,
+    ) -> Result<Self> {
         let root_path = MetadataPath::from_role(&Role::Root);
+        let (local, remote) = (local.into(), remote.into());
```
erickt: Small nit, I recommend doing the `.into()` first, since logically the `root_path` and `root_version` are more closely related than this `(local, remote)`.

src/client.rs
```diff
@@ -356,6 +371,10 @@ where
         )
         .await?;
 
+        // FIXME why does with_trusted_root_keyids store using the metadata version but
+        // with_trusted_root_keys use the version from the caller?
+        let root_version = MetadataVersion::Number(trusted_root.version());
```
erickt: I think I just forgot to copy this case into `with_truted_root_keys`. We haven't gotten any checks yet that the metadata returned the version we requested (see #253), so the safest thing is to use the metadata version, since the whole system could get rather confused if we stored it with the wrong version.

src/repository.rs
```diff
@@ -82,57 +175,334 @@ where
...
+    ) -> Result<(RawSignedMetadata<D, M>, SignedMetadata<D, M>)>
     where
-        M: Metadata + Sync + 'static,
+        M: Metadata,
     {
-        (**self).store_metadata(meta_path, version, metadata)
+        let raw_signed_meta = self
+            .fetch_raw_metadata(meta_path, version, max_length, hash_data)
+            .await?;
+        let signed_meta = raw_signed_meta.parse()?;
```
erickt, line 205: might be better to move this to the caller, to make it easier to see where the parsing happens?

src/util.rs
```diff
@@ -1,32 +1,118 @@
...
+        self,
+        max_length: u64,
+        hash_data: Option<(&HashAlgorithm, HashValue)>,
+    ) -> Result<SafeReader<Self>> {
+        SafeReader::new(self, max_length, hash_data)
+    }
+}
+
+/// Wraps an `AsyncRead` to detect and fail transfers slower than a minimum bitrate.
+pub(crate) struct EnforceMinimumBitrate<R> {
```
erickt: This needs a test.

src/repository/http.rs
```diff
@@ -243,94 +230,56 @@ where
...
-        target_description: &'a TargetDescription,
+        _target_description: &'a TargetDescription,
     ) -> BoxFuture<'a, Result<Box<dyn AsyncRead + Send + Unpin>>> {
         async move {
-            let (alg, value) = crypto::hash_preference(target_description.hashes())?;
             let components = target_path.components();
             let resp = self.get(&self.targets_prefix, &components).await?;
 
-            let stream = resp
+            // TODO check content length if known and fail early if the payload is too large.
```
erickt: Can you file a ticket for this, and link to it here?

src/repository.rs
```diff
@@ -19,53 +23,142 @@ pub use self::http::{HttpRepository, HttpRepositoryBuilder};
...
 
-/// Top-level trait that represents a TUF repository and contains all the ways it can be interacted
-/// with.
-pub trait Repository<D>
-where
-    D: DataInterchange + Sync,
-{
-    /// Store signed metadata.
+/// A readable TUF repository.
+pub trait RepositoryProvider {
```
erickt: Why did you move the `DataInterchange` parameter onto the methods? This has the side effect of us not being able to create a `Box<dyn RepositoryProvider>`.

I suppose this would allow us to interact with a repository that supports interchanges, but I'm not sure if that's worth trying to support just yet, since we (and the tuf spec) hasn't actually figured out how alternative interchanges would work.

tests/simple_example.rs
```diff
@@ -105,7 +105,7 @@ where
 }
 
 async fn init_server<'a, T>(
-    remote: &'a mut EphemeralRepository<Json>,
+    remote: &'a mut Repository<EphemeralRepository, Json>,
```
erickt: What do you think about just passing in `EphemeralRepository` and not wrap it in `Repository`? The only thing that buys us is that we call `Repository::check` to see if the metadata path matches the role, but I'm not sure if that check actually buys us that much. If a server messed up and wrote the root metadata as `timestamp.json`, the client would fetch that file then it should fail to deserialize into `TimestampMetadata`. So really, this check is more for making sure we don't generate a malformed repository, but given how low-level the current implementation is, the whole thing probably needs to be architected (see #232).

src/repository.rs
```diff
@@ -19,53 +23,142 @@ pub use self::http::{HttpRepository, HttpRepositoryBuilder};
...
+    {
+        (**self).store_target(target, target_path)
+    }
+}
+
+/// A wrapper around an implementation of [`RepositoryProvider`] and/or [`RepositoryStorage`] tied
+/// to a specific [`DataInterchange`](crate::interchange::DataInterchange) that will enforce
+/// provided length limits and hash checks.
+#[derive(Debug, Clone)]
+pub struct Repository<R, D> {
```
erickt: I think we could make this `pub(crate)`. It's mainly about making sure the client doesn't accidentally access the repositories directly. I'm not sure if external users would need this.

src/repository.rs
```diff
@@ -19,53 +23,142 @@ pub use self::http::{HttpRepository, HttpRepositoryBuilder};
...
+        target_path: &'a TargetPath,
+        target_description: &'a TargetDescription,
+    ) -> BoxFuture<'a, Result<Box<dyn AsyncRead + Send + Unpin>>>;
+}
 
-    /// Fetch signed metadata.
-    fn fetch_metadata<'a, M>(
+/// A writable TUF repository. Most implementors of this trait should also implement
+/// `RepositoryProvider`.
+pub trait RepositoryStorage {
```
erickt: As above, I think this should take `D` as a generic parameter.

erickt (changes requested):

src/util.rs
wellsie1116: https://github.com/heartsucker/rust-tuf/pull/277

src/repository.rs
wellsie1116: Because `Repository<EphemeralRepository<Json>, Json>` was redundant, and the repository provider is where the metadata and targets are, in whatever form they take.  Yeah, it is a bit weird to allow a provider to provide different data interchanges when we don't have a story for what that looks like yet.

I'll plumb the type parameter back through the trait itself.

src/repository.rs
wellsie1116: Fine with me.  It was a little more convenient to use in the test binaries in this crate, but it isn't too terrible to use the repository implementations directly, and I suppose I could add some similar methods to EphemeralRepository to simplify testing if we find ourselves doing a lot of repository manipulation from external tests.

tests/simple_example.rs
wellsie1116: Not quite.  The reason I used Repository in these test binaries was because its method signatures were nicer to use (I could store RawMetadata instead of having to convert them to bytes first).

src/client.rs
wellsie1116: Removed Repository from the public interface, so these are all back to L and R.

src/client.rs
wellsie1116: Added that to with_trusted_root_keys then, thanks for clarifying.

src/repository/http.rs
wellsie1116: https://github.com/heartsucker/rust-tuf/issues/278 (also added to the TODO)

tests/interop/main.rs
```diff
@@ -206,10 +207,16 @@ async fn extract_keys(dir: &Path) -> Vec<KeyId> {
     let remote = init_remote(dir).unwrap();
 
     let root_path = MetadataPath::from_role(&Role::Root);
-    let metadata: SignedMetadata<_, RootMetadata> = remote
+
+    let mut buf = Vec::new();
```
wellsie1116, line 211: This block is a good example of why I exported `Repository`.

src/repository.rs
wellsie1116: The match statements in lookup_target_description get rather nasty if there isn't a single method I can call to get both.

wellsie1116:
Merged in develop and responded to the comments. Please take another look.

src/repository.rs
erickt: I've got a little concern that if we ever have two `fetch_metadata` calls in the same scope, it'd be a little easier to hash check one metadata, but use the other when trying to update the tuf database. This probably isn't a large problem in practice since we still go through signature verification and the rest of the validation, but it'd be nice to avoid this.

We can defer addressing this for a future patch though.

erickt (approved):

PR #275, opened 2020-02-04
Verify top-level delegated targets signatures

wellsie1116:
The previous implementation did not consider delegated roles from the
top-level "targets" role when looking for keys to verify the delegated
targets with, which would result in delegated targets being accepted by
Tuf without signature verification.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -782,7 +782,7 @@ where
         current_depth: u32,
         target: &'a VirtualTargetPath,
         snapshot: &'a SnapshotMetadata,
-        targets: Option<&'a TargetsMetadata>,
+        targets: Option<(&'a TargetsMetadata, MetadataPath)>,
     ) -> (bool, Result<TargetDescription>) {
         if current_depth > self.config.max_delegation_depth {
             warn!(
@@ -794,10 +794,10 @@ where
 
         // these clones are dumb, but we need immutable values and not references for update
         // tuf in the loop below
-        let targets = match targets {
-            Some(t) => t.clone(),
+        let (targets, targets_role) = match targets {
+            Some((t, role)) => (t.clone(), role),
             None => match self.tuf.targets() {
-                Some(t) => t.clone(),
+                Some(t) => (t.clone(), MetadataPath::from_role(&Role::Targets)),
                 None => {
                     return (
                         default_terminate,
@@ -881,7 +881,7 @@ where
 
             match self
                 .tuf
-                .update_delegation(delegation.role(), signed_meta.clone())
+                .update_delegation(&targets_role, delegation.role(), signed_meta.clone())
             {
                 Ok(_) => {
                     match self
@@ -909,7 +909,7 @@ where
                             current_depth + 1,
                             target,
                             snapshot,
-                            Some(meta.as_ref()),
+                            Some((meta.as_ref(), delegation.role().clone())),
                         ));
                     let (term, res) = f.await;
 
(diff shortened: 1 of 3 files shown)


src/tuf.rs
```diff
@@ -438,27 +479,23 @@ impl<D: DataInterchange> Tuf<D> {
                 return Ok(false);
             }
 
-            for delegated_targets in self.delegations.values() {
-                let parent = match delegated_targets.as_ref().delegations() {
-                    Some(d) => d,
-                    None => &targets_delegations,
-                };
+            // FIXME Since the delegating targets defines the valid keys a delegated targets can be
```
erickt: Can you file a ticket for this fixme?

src/tuf.rs
```diff
@@ -438,27 +479,23 @@ impl<D: DataInterchange> Tuf<D> {
...
+            // signed by, it is possible for delegated targets to be valid when accessed via one
+            // target path and invalid when accessed via another. Instead of verifying all known
+            // references to this delegation, this verification step should be choosing just the
+            // one for the target path the outer tuf::Client is trying to resolve, but that
+            // currently depends on state in tuf::Client to do so.
```
erickt: Can you file a ticket for this fixme?

Ideally `Tuf` wouldn't have to trust outside state from `tuf::Client`. For the short term though, what do you think about `update_delegation` also getting passed in the parent `Role` name? Then we could directly look that up in our map and use that to verify this delegation. That should allow us to support the whole "this delegation is valid when found through this delegation chain, but not valid when found through that delegation chain".

src/tuf.rs
```diff
@@ -395,6 +395,50 @@ impl<D: DataInterchange> Tuf<D> {
         Ok(true)
     }
 
+    /// Walk the base targets delegations and any known nested delegations for the authorized
+    /// signing keys and metadata for the delegation given by `role`.
+    fn find_delegation(&self, role: &MetadataPath) -> Vec<(Vec<&PublicKey>, &Delegation)> {
```
erickt: Good old tree searching. Unfortunately delegations could potentially have a loop, so can you add in a "I visited this node already" set to avoid infinite loops?

erickt (changes requested):
Finally, we really need some test coverage here. Would you want to add it? Or we could add a "delegation" feature flag to compile out support until it's ready?

src/tuf.rs
wellsie1116: #279

src/tuf.rs
wellsie1116: That is a good idea for a short term fix.  Implemented.

src/tuf.rs
wellsie1116: Of course, why wouldn't that be supported :)

Fixed by finding just the delegation we are looking for without having to walk the tree.

wellsie1116:
There were already some integration style tests in integration.rs, which is how I found this issue in the first place.  I have extended them to:
 * verify signatures of the first level of delegated targets are being verified (and verified that test used to fail), and
 * verify that delegations can be considered valid or invalid, depending on the path to them.

src/client.rs
```diff
@@ -909,7 +909,7 @@ where
                             current_depth + 1,
                             target,
                             snapshot,
-                            Some(meta.as_ref()),
+                            Some((meta.as_ref(), delegation.role().clone())),
```
wellsie1116, line 912: Am I using the right role here?  I think so, but the integration test also wouldn't catch a mistake here.

erickt: Yeah I think this is the right role.

erickt (approved):

PR #276, opened 2020-02-05
Split SafeReader into separate types

wellsie1116:
Since the Repository traits can be implemented outside this crate, it is
not safe to assume that those implementations will verify the maximum
length or hash of metadata/targets before succeeding a fetch request.

In preparation to move the length check and hash verification into the
tuf client, this change splits the SafeReader into 2 types:
 * EnforceMinimumBitrate enforces a minimum transfer rate, currently
   utilized by only the http repository implementation
 * SafeReader, which retains the logic to enforce a maximum file size
   and hash value.
This change also defines an extension trait on AsyncRead to easily wrap
an AsyncRead in these types.

A future change will move hash and length enforcement out of the
Repository implementations.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -1,5 +1,3 @@
-use chrono::offset::Utc;
-use chrono::DateTime;
 use futures_io::AsyncRead;
 use futures_util::ready;
 use ring::digest;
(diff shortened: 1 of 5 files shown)


src/util.rs
```diff
@@ -1,32 +1,118 @@
-use chrono::offset::Utc;
```
erickt, line 1: Can we remove the chrono dependency?

src/util.rs
```diff
@@ -1,32 +1,118 @@
...
 use std::io::{self, ErrorKind};
 use std::marker::Unpin;
 use std::pin::Pin;
 use std::task::{Context, Poll};
+use std::time::{Duration, Instant};
 
 use crate::crypto::{HashAlgorithm, HashValue};
 use crate::Result;
 
+pub(crate) trait SafeAsyncRead: Sized {
```
erickt, line 13: It's probably fine that this is `pub(crate)` for now, but I can imagine custom `Repository` implementations might want to use this functionality.

erickt (approved):

src/util.rs
erickt: nope, it's used elsewhere.

Issue #136, opened 2018-07-17
Use `failure` crate for errors instead of a enum


heartsucker:

wellsie1116:
Are there any concerns with tuf::Error not being able to implement PartialEq/Eq?  I'd like to extract some of the error cases currently stored in Error::Opaque into their own variants, but errors like `failure::Error`, `anyhow::Error`, `hyper::Error` and `std::io::Error`, for example, aren't PartialEq.

This can be a problem with `assert_eq!` in tests, but `assert_matches!` mostly solves that issue.

PR #280, opened 2020-02-06
Derive Error implementation

wellsie1116:
Replace the manual implementation of std::error::Error for Error with the thiserror::Error derive macro. Other than the error strings containing more information and thiserror::Error implementing std::error::Error::description() with a static string indicating the method is deprecated, no observable changes expected.

Related to #136

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ serde_json = "1"
 tempfile = "3"
 untrusted = "0.7"
 url = "2"
+thiserror = "1.0"
 
 [dev-dependencies]
 lazy_static = "1"
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,74 +1,68 @@
 //! Error types and converters.
 
 use data_encoding::DecodeError;
-use derp;
-use http;
-use hyper;
-use serde_json;
-use std::fmt;
 use std::io;
 use std::path::Path;
-use tempfile;
+use thiserror::Error;
 
 use crate::metadata::Role;
 
 /// Error type for all TUF related errors.
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Error, Debug, PartialEq, Eq)]
 pub enum Error {
     /// The metadata had a bad signature.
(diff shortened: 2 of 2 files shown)


heartsucker:
I'm not sure how this is more useful than `failure` or why this should be picked. I have used `failure` in other places and been happy with the way that API works.. I also don't see much of a change with this compared to the current impl. What's the benefit here?

wellsie1116:
First off, I am 100% okay with using failure.  I too have used it and like its API surface and derive macro.

thiserror/anyhow are very similar, except:
* they are based on the std::error::Error trait, which is less broken these days, removing some of the incentive to depend on or export a custom trait
* they are at 1.0, failure has no plans to stabilize
* they produce nicer error output in my opinion
* thiserror allows fields to be tagged with #[from] to auto-generate From<> conversion to that error variant

This pull request as-is doesn't change much.  The Display implementation will include the context from the error variants now and the "soft-deprecated" display method is now a stub, but that is about it.

The reason why I'm poking at the `Error` type at all is that I am interested in teasing apart some of the cases in Error::Opaque.  Specifically, I'd like to know if a Client::update failed due to a network related error.  But if I have an error variant that wraps a `hyper::Error`, then I should probably implement `std::error::Error::source`, and if I'm going to implement `source`, I might as well switch to a derive helper, so here we are.

So, my understanding of thiserror/anyhow is it does roughly everything failure does in a subjectively prettier and more stable way, but again, I'm fine with using either.

Also, hi :)

erickt (approved):
I like this too, but I'll wait for @heartsucker to approve before merging since they haven't responded to your response yet.

heartsucker:
I think this is reasonable, and at the time I wrote this, I didn't have a way to nicely include more detailed context in the `Opaque`. I think error handling in this could be overall improved, and I know that `failure` has some modes of usage that allow for boxing and downcasting as "pattern matching" which would allow for some extraction.

I agree that this is an improvement, and in theory we could improve the overall structure under here too.

PR #282, opened 2020-02-14
Signature preserves unknown fields

### LINKED ISSUE #223: rust-tuf should preserve unknown fields in metadata
While the TUF spec explicitly calls out some areas where metadata can be customized (like the target `custom` field), it does not forbid metadata from containing other non-standard fields. For example, python-tuf adds a `keyid_hash_algorithms` field to the root key structure, which is incorporated into that key's key id. rust-tuf currently deserializes metadata into a rust struct, then re-serializes it into cjson in order to validate metadata. Unfortunately though this means that we throw away unknown fields, and therefore will fail to validate the signature.

To fix this, I think we have two options:

1. Rather than directly deserializing some signed metadata like the root metadata into:

```
#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
pub struct SignedMetadata<D, M> {
    signatures: Vec<Signature>,
    #[serde(rename = "signed")]
    metadata: M,
    #[serde(skip_serializing, skip_deserializing)]
    _interchage: PhantomData<D>,
}
```

Instead, we should deserialize into a type that stores the signed metadata in a `serde_json::Object`, which can represent any json value. So something like this:

```
#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
pub struct SignedMetadata<D, M> {
    signatures: Vec<Signature>,
    signed: serde_json::Value,
    #[serde(skip_serializing, skip_deserializing)]
    metadata: Option<M>,
    #[serde(skip_serializing, skip_deserializing)]
    _interchage: PhantomData<D>,
}
```

This structure then is what is serialized to disk, and the `signed` portion is used to validate the signature. Once validated, the `signed` portion can be deserialized into the in memory `metadata` field for further use.

2. We could annotate all structs with [`#[serde(flatten)]`](https://serde.rs/attr-flatten.html) to capture unknown values into a side structure. This then should preserve those values in order to pass validation, but has the side effect of adding a lot of fields, and it might be easy to forget to capture or propagate these fields everywhere, so I think option 1 might be a more robust implementation.

wellsie1116:
Per the TUF spec, all metadata formats "include the ability to add more attribute-value fields for backwards-compatible format changes." So, if the signed portion of some metadata contains unexpected fields, those fields should not be ignored until after verifying the signature of the metdata.

These changes:
 * replace the parsed metadata in the SignedMetadata struct with a D::RawData, which will preserve unknown fields.
 * changes the verify method on SignedMetadata to produce the fully parsed metadata without any attached signatures, as those are no longer able to verify the parsed metadata.
 * verify that signatures of metadata containing unknown fields verify correctly.

Fixed: #223

### DIFF
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -269,14 +269,17 @@ where
     }
 }
 
-/// A piece of metadata with attached signatures.
+/// Serialized metadata with attached unverified signatures.
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
-pub struct SignedMetadata<D, M> {
+pub struct SignedMetadata<D, M>
+where
+    D: DataInterchange,
+{
     signatures: Vec<Signature>,
     #[serde(rename = "signed")]
-    metadata: M,
+    metadata: D::RawData,
     #[serde(skip_serializing, skip_deserializing)]
-    _interchage: PhantomData<D>,
+    _marker: PhantomData<M>,
 }
 
 impl<D, M> SignedMetadata<D, M>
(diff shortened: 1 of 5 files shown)


src/metadata.rs
```diff
@@ -501,38 +573,54 @@ where
...
+    /// latest root.json.
+    pub fn untrusted_info(&self) -> Result<MetadataMetadata<RootMetadata>> {
+        D::deserialize(&self.metadata)
     }
 }
 
-impl<D, M> Metadata for SignedMetadata<D, M>
+/// Metadata common to all signed metadata files.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct MetadataMetadata<M> {
```
erickt: This doesn't need to be public either.

I don't think we need this either. We could instead make `parse_version_untrusted` a `pub(crate)` to extract out the version number.

src/metadata.rs
```diff
@@ -501,38 +573,54 @@ where
...
+impl<D> SignedMetadata<D, RootMetadata>
+where
+    D: DataInterchange,
+{
+    /// Parse common metadata fields from this metadata without verifying signatures.
+    ///
+    /// This operation is generally unsafe to do with metadata obtained from an untrusted source,
+    /// but rolling forward to the most recent root.json requires using the version number of the
+    /// latest root.json.
+    pub fn untrusted_info(&self) -> Result<MetadataMetadata<RootMetadata>> {
```
erickt: I don't think this needs to be public, you didn't use it in any of the tests.

erickt (changes requested):

src/metadata.rs
wellsie1116: fixed.

src/metadata.rs
wellsie1116: fixed.

erickt (approved):

PR #284, opened 2020-03-24
Switch to using the serde derive feature

erickt:
The serde docs now recommend using the `derive` feature over using `serde_derive`.


ComputerDruid (approved):

erickt:
Decided against this since there wasn't that much benefit.

PR #285, opened 2020-04-15
Add functions to get current trusted versions of metadata.

wittrock:
Will be useful for debugging client behavior.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -542,6 +542,31 @@ where
         }
     }
 
+    /// Returns the current trusted root version.
+    pub fn root_version(&self) -> u32 {
+        self.tuf.root().version()
+    }
+
+    /// Returns the current trusted timestamp version.
+    pub fn timestamp_version(&self) -> Option<u32> {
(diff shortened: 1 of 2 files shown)


src/client.rs
```diff
@@ -1646,6 +1671,12 @@ mod test {
         assert_eq!(client.update().await, Ok(true));
         assert_eq!(client.tuf.root().version(), 2);
 
+        assert_eq!(client.root_version(), 2);
+        assert_eq!(client.timestamp_version(), Some(1));
+        assert_eq!(client.snapshot_version(), Some(1));
+        assert_eq!(client.targets_version(), Some(1));
+        assert_eq!(client.delegations_version(&snapshot_path), None);
```
erickt, line 1678: Can you add a test to check `delegations_version` returns a value?

erickt (commented):
Other than that, looks good to me.

erickt:
There isn't a convenient test to bootstrap this on, so I'm going to land it now, and add a better test for this in a new PR.

PR #287, opened 2020-04-15
Add RepoBuilder metadata generator, and tests

erickt:
This is a stack of a few patches. First, it introduces `RepoBuilder`, a helper builder that takes away some of the boilerplate for writing metadata. It's not complete though, so I've left it private and only exposed to tests. It needs the following fixed before I think we should expose it:

* Fetch and validate the old metadata before basing new metadata off of it.
* Don't overwrite metadata that already exists.
* Auto-increment the version numbers.
* Optionally allow the timestamp version number to be generated off of the current UTC time, to ease working with test repositories.
* Merge targets and delegates.
* Come up with a better way to handle delegation metadata.
* Add documentation and examples.
* Make a proper error for when `repo_builder::sign` is given an empty vec of private keys.

In addition, it adds and extends a of tests to use it, and finally, I found and fixed a bug that prevented delegated target descriptions from being fetched. I added a test for this as well to make sure it keeps working.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1145,10 +1145,8 @@ mod test {
     use super::*;
     use crate::crypto::{HashAlgorithm, KeyType, PrivateKey, SignatureScheme};
     use crate::interchange::Json;
-    use crate::metadata::{
-        MetadataPath, MetadataVersion, RootMetadata, RootMetadataBuilder, SnapshotMetadataBuilder,
-        TargetsMetadataBuilder, TimestampMetadataBuilder,
-    };
+    use crate::metadata::{MetadataPath, MetadataVersion, RootMetadata, RootMetadataBuilder};
+    use crate::repo_builder::{RepoBuilder, RepoKeys};
     use crate::repository::EphemeralRepository;
     use chrono::prelude::*;
     use futures_executor::block_on;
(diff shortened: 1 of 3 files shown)


src/client.rs
```diff
@@ -952,13 +952,15 @@ where
                             snapshot,
                             Some((&meta, delegation.role().clone())),
                         ));
-                    let (term, res) = f.await;
 
-                    if term && res.is_err() {
-                        return (true, res);
+                    match f.await {
+                        (term, res @ Ok(_)) => return (term, res),
```
wellsie1116: by passing `term` through to the caller, it is now unused.  Something seems off with that.

src/client.rs
```diff
@@ -1468,35 +1345,29 @@ mod test {
...
+                .commit()
                 .await
                 .unwrap();
 
-            client
-                .remote
-                .store_metadata(&root_path, &MetadataVersion::None, &root3.to_raw().unwrap())
+            // Make sure the version 3 is signed by version 2's keys.
+            let repo_keys3 = RepoKeys {
+                root_keys: vec![&KEYS[0], &KEYS[1]],
```
wellsie1116: should be keys 1 and 2.

src/client.rs
```diff
@@ -1468,35 +1345,29 @@ mod test {
...
+            // Make sure the version 2 is signed by version 1's keys.
+            let repo_keys2 = RepoKeys {
+                root_keys: vec![&KEYS[0], &KEYS[1]],
+                snapshot_keys: vec![&KEYS[0], &KEYS[1], &KEYS[2]],
+                targets_keys: vec![&KEYS[0], &KEYS[1], &KEYS[2]],
+                timestamp_keys: vec![&KEYS[0], &KEYS[1], &KEYS[2]],
+            };
+            let _root2 = RepoBuilder::new(repo_keys2, &repo)
+                .with_root(|b| b.version(2).expires(Utc.ymd(2038, 1, 1).and_hms(0, 0, 0)))
+                .commit()
```
wellsie1116, line 1216: the test previously kept the same targets/snapshot/timestamp metadata and just regenerated the root metadata, but those are being re-built now.  I'm not sure if signatures are guaranteed to be the same if all the inputs are the same.  Timestamp would still indicate that snapshot version 1 is the most recent, so the client wouldn't notice clobbered remote metadata.

src/client.rs
```diff
@@ -1535,117 +1406,35 @@ mod test {
...
-            )
-            .await
-            .unwrap();
-
-        remote
-            .store_metadata(&root_path, &MetadataVersion::None, &root1.to_raw().unwrap())
-            .await
-            .unwrap();
-
-        let metadata_version = if consistent_snapshot {
```
wellsie1116, line 1467: is a repository in consistent_snapshots mode supposed to write unversioned targets/snapshots?  A functional change with RepoBuilder is that it always writes the unversioned files, but I'm not sure if that's a bad thing.  I suppose if the client were to start incorrectly reading the unversioned files, this test wouldn't catch it anymore.

src/repo_builder.rs
```diff
@@ -0,0 +1,249 @@
...
+};
+use crate::repository::{Repository, RepositoryProvider, RepositoryStorage};
+use crate::Result;
+
+// This is a helper crate to keep track of the private keys necessary to create new metadata.
+//
+// FIXME: This is not ready yet for public use, it is only intended for internal testing until the
+// design is complete.
+pub(crate) struct RepoKeys<'a> {
+    pub(crate) root_keys: Vec<&'a PrivateKey>,
```
wellsie1116: nit: these probably don't need a "_keys" suffix.

src/repo_builder.rs
```diff
@@ -0,0 +1,249 @@
...
+            .await?;
+
+        self.repo
+            .store_metadata(&root_path, &MetadataVersion::None, &root.to_raw()?)
+            .await?;
+
+        // Commit the delegated targets metadata.
+        for (path, delegated_targets) in self.delegated_targets {
+            self.repo
+                .store_metadata(&path, &MetadataVersion::None, &delegated_targets.to_raw()?)
```
wellsie1116: delegated targets aren't also written to versioned paths?

wellsie1116 (commented):
Defining multiple sets of TUF metadata in a test is rather verbose, so I like the idea of a RepoBuilder, and this is a great start on one.  I added a few questions/nits.

I need to give the delegations section of the spec a good long read.

src/repo_builder.rs
erickt: oops you are right.

src/client.rs
erickt: Technically the spec makes it optional, but you are right, it's probably not worth us doing. I'll fix it.

src/client.rs
erickt: done

src/client.rs
erickt: I'll just switch to `(_, res @ Ok(_)) => return (true, res)` so that we explicitly terminate if we ever find our target.

src/client.rs
erickt: That's a good point. I believe it depends on the signature algorithm. I think ed25519 signatures are consistent, but rsa signatures are (or can be) non-deterministic.

However, it looks like it might be a little complicated to clean things up to support that. Would you be okay if we address this in a future PR?

erickt:
Superseded by #307 (and some other patches that haven't been published yet).

Issue #286, opened 2020-04-15
Fix spelling for UnkonwnHashAlgorithm variant


zkbaum:
The current spelling is `UnkonwnHashAlgorithm`. It should be `UnknownHashAlgorithm`.

https://github.com/heartsucker/rust-tuf/blob/62431b1d4dc45fef8c63d900b9b668af5fc33faf/src/error.rs#L57

erickt:
Fixed in #288

PR #288, opened 2020-06-01
Adding Hyper and Http error variants. 

itsmeolivia:
Also,removes UnknownHashAlgorithm

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -8,6 +8,7 @@ use thiserror::Error;
 use crate::metadata::Role;
 
 /// Error type for all TUF related errors.
+#[non_exhaustive]
 #[derive(Error, Debug)]
 pub enum Error {
     /// The metadata had a bad signature.
--- a/src/error.rs
+++ b/src/error.rs
@@ -9,7 +9,7 @@ use crate::metadata::Role;
 
 /// Error type for all TUF related errors.
 #[non_exhaustive]
-#[derive(Error, Debug, PartialEq, Eq)]
+#[derive(Error, Debug)]
 pub enum Error {
     /// The metadata had a bad signature.
     #[error("bad signature")]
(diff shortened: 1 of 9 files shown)


src/error.rs
```diff
@@ -54,7 +62,7 @@ pub enum Error {
 
     /// There is no known or available hash algorithm.
     #[error("unknown hash algorithm: {0}")]
-    UnkonwnHashAlgorithm(String),
+    UnknownHashAlgorithm(String),
```
erickt: It seems like this isn't used. Could you just remove it?

src/error.rs
```diff
@@ -26,6 +26,14 @@ pub enum Error {
     #[error("illegal argument: {0}")]
     IllegalArgument(String),
 
+    /// Generic error for HTTP connections.
+    #[error("http: {0}")]
+    Http(String),
```
erickt: Why not propagate the error from the `http` crate?

src/error.rs
```diff
@@ -26,6 +26,14 @@ pub enum Error {
...
+
+    /// Errors that can occur parsing HTTP streams.
+    #[error("hyper: {0}")]
+    Hyper(String),
```
erickt: Here too.

erickt (commented):
Since you are making a breaking change to the `Error` type, could you also mark this type as `#[non_exhaustive]`, as described by https://doc.rust-lang.org/1.35.0/unstable-book/language-features/non-exhaustive.html? This will allow us to add new error types in the future.

src/error.rs
itsmeolivia: http::Error and hyper::Error do not implement PartialEq or Eq

itsmeolivia:
for the non_exhaustive tag, the ci needs to be updated to 1.40.0

itsmeolivia:
closes #286

src/error.rs
erickt: Do we still need `PartialEq` and `Eq` if we're using `assert_matches`? Are there other places we need equality?

erickt (changes requested):

PR #289, opened 2020-09-17
Remove tuf::Client::with_trusted_root_keyids

erickt:
In https://github.com/theupdateframework/taps/blob/master/tap12.md, the
TUF community is considering removing the requirement that the keyid must
be derived from the public key. If this lands, then it wouldn't be safe to
initialize trust with just the keyids. To prepare for this transition,
this patch removes `tuf::Client::with_trusted_root_keyids`. If this
functionality is still needed, it'd be relatively straightforward for a
client to re-write this method in their codebase.

In help implement this, this patch also extends RootMetadata to add helper
methods to extract out role public keys.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -8,24 +8,21 @@
 //! # use std::path::PathBuf;
 //! # use std::str::FromStr;
 //! # use tuf::{Result, Tuf};
-//! # use tuf::crypto::KeyId;
+//! # use tuf::crypto::PublicKey;
 //! # use tuf::client::{Client, Config};
 //! # use tuf::metadata::{RootMetadata, Role, MetadataPath, MetadataVersion};
 //! # use tuf::interchange::Json;
 //! # use tuf::repository::{FileSystemRepository, HttpRepositoryBuilder};
-//!
-//! static TRUSTED_ROOT_KEY_IDS: &'static [&str] = &[
-//!     "4750eaf6878740780d6f97b12dbad079fb012bec88c78de2c380add56d3f51db",
-//!     "01892c662c8cd79fab20edec21de1dcb8b75d9353103face7fe086ff5c0098e4",
-//!     "0823260c29956d0b64baa19a64c49c87922ba2bc532d09a6965aef044da490bd",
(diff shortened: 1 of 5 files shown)


src/client.rs
```diff
@@ -8,23 +8,17 @@
...
 //! # use tuf::metadata::{RootMetadata, Role, MetadataPath, MetadataVersion};
 //! # use tuf::interchange::Json;
 //! # use tuf::repository::{FileSystemRepository, HttpRepositoryBuilder};
 //!
-//! static TRUSTED_ROOT_KEY_IDS: &'static [&str] = &[
-//!     "4750eaf6878740780d6f97b12dbad079fb012bec88c78de2c380add56d3f51db",
-//!     "01892c662c8cd79fab20edec21de1dcb8b75d9353103face7fe086ff5c0098e4",
-//!     "0823260c29956d0b64baa19a64c49c87922ba2bc532d09a6965aef044da490bd",
-//! ];
+//! static TRUSTED_ROOT_KEYS: &'static [PublicKey] = &[];
```
wellsie1116: For this top-level crate documentation, consider including some PublicKey::new calls here (might need to make that a const function), or defer to a hidden "load_trusted_public_keys()" function to get them.  Threshold is still 1 below, so this wouldn't be a very useful client as-is.

src/client.rs
```diff
@@ -8,23 +8,17 @@
...
 //!
 //! # fn main() -> Result<()> {
 //! # block_on(async {
-//! let key_ids: Vec<KeyId> = TRUSTED_ROOT_KEY_IDS.iter()
-//!     .map(|k| KeyId::from_str(k).unwrap())
-//!     .collect();
+//! let keys: Vec<PublicKey> = TRUSTED_ROOT_KEYS.iter().cloned().collect();
```
wellsie1116: with_trusted_root_keys wants an IntoIterator of &PublicKey, which the slice impls directly, so this logic is no longer necessary.

wellsie1116 (approved):

PR #290, opened 2020-09-17
Rename Tuf methods and variables

erickt:
There's a risk when reviewing `tuf.rs`'s code that one could mistake the trusted metadata with the untrusted metadata we are still validating. This adds:

* `trusted_` prefix to all the instances where we are deriving data from the trusted metadata.
* `new_` to all the instances where we are working with the new, untrusted metadata.
* renamed `safe_$ROLE_ref()` to `trusted_$ROLE_unexpired()` to make it clear it will err if the metadata is expired.

This makes it easier to check if the metadata is being used incorrectly.

Test: This doesn't have any functional changes, so it doesn't need tests.

### DIFF
--- a/src/tuf.rs
+++ b/src/tuf.rs
@@ -180,7 +180,7 @@ impl<D: DataInterchange> Tuf<D> {
     /// Returns a reference to the parsed metadata if the metadata was newer.
     pub fn update_timestamp(
         &mut self,
-        signed_timestamp: SignedMetadata<D, TimestampMetadata>,
+        new_signed_timestamp: SignedMetadata<D, TimestampMetadata>,
     ) -> Result<Option<&TimestampMetadata>> {
         let verified = {
             // FIXME(https://github.com/theupdateframework/specification/issues/113) Should we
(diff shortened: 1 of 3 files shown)


src/tuf.rs
```diff
@@ -9,19 +9,19 @@ use crate::crypto::PublicKey;
...
 use crate::interchange::DataInterchange;
 use crate::metadata::{
-    Delegation, Delegations, Metadata, MetadataPath, Role, RootMetadata, SignedMetadata,
-    SnapshotMetadata, TargetDescription, TargetsMetadata, TimestampMetadata, VirtualTargetPath,
+    Delegations, Metadata, MetadataPath, Role, RootMetadata, SignedMetadata, SnapshotMetadata,
+    TargetDescription, TargetsMetadata, TimestampMetadata, VirtualTargetPath,
 };
 use crate::Result;
 
 /// Contains trusted TUF metadata and can be used to verify other metadata and targets.
```
wellsie1116, line 17: it seems a bit redundant to have `trusted_` prefixes on the fields themselves since `Tuf` only ever stores trusted metadata.  I'm fine with either though.

src/tuf.rs
```diff
@@ -171,13 +183,16 @@ impl<D: DataInterchange> Tuf<D> {
         signed_timestamp: SignedMetadata<D, TimestampMetadata>,
```
wellsie1116: Consider renaming signed_timestamp (plus the other metadata types) to have a `new_` prefix too. This may or may not be signed correctly.

wellsie1116 (approved):

src/tuf.rs
erickt: Agree it's quite redundant. However, since the `Tuf` type is the thing that manages the security properties of rust-tuf, I figured it couldn't hurt to have additional signal. That's also why I wrapped these types in a `Verified` struct so that we can get the compiler to enforce we're using the right thing.

Issue #291, opened 2020-09-17
rust-tuf should not deserialize metadata until it has been verified


erickt:
It's possible that a parser bug could allow for a man-in-the-middle attacker to compromise a system. None have been found so far in serde_json, but other parsers have had bugs like this, see [CVE-2017-18349](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-18349). We would like to modify rust-tuf to only deserialize metadata after it has been verified, either by hash or signature.

Unfortunately, there are a few challenges here.

* [POUF-1] embeds the signatures into the role metadata.
* The TUF-1.0.5 spec itself implies that metadata is parsed before validation in sections 5.3.1 and 5.4.1 in order to verify metadata versions before signatures. In https://github.com/theupdateframework/specification/pull/112, I've proposed moving this check to after signature verification.
* rust-tuf isn't careful on when deserialization happens.

We should address these to protect ourselves from these attacks.

[POUF-1]: https://github.com/theupdateframework/taps/blob/master/POUFs/reference-POUF/pouf1.md

PR #293, opened 2020-09-17
Control when metadata is deserialized

erickt:
This rewrites rust-tuf to be more strict on when metadata is deserialized. This is to start laying the groundwork to prevent rust-tuf from being vulnerable to parser exploits (like [CVE-2017-18349]). Almost all deserialization is moved into the new `verify.rs` module, which validates the metadata's signature before returning the deserialized metadata to the caller.

There are two remaining cases where metadata is parsed outside of `verify`:

* `Client::from_trusted_local`, where we don't have any keys to establish trust of the initial root metadata.
* `Client::update_root`, where we fetch the latest root metadata, and walk from the current trusted metadata version to the latest root metadata version.

These are explicitly called out with WARNING comments.

Also, this starts incorporating the TUF workflow spec into the `Tuf` functionality, in order to help reviewers understand which code is implementing which part of the spec, or make it obvious when we diverge from it.

Bug: #291

[CVE-2017-18349]: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-18349

### DIFF
--- a/src/tuf.rs
+++ b/src/tuf.rs
@@ -9,19 +9,20 @@ use crate::crypto::PublicKey;
 use crate::error::Error;
 use crate::interchange::DataInterchange;
 use crate::metadata::{
-    Delegations, Metadata, MetadataPath, Role, RootMetadata, SignedMetadata, SnapshotMetadata,
-    TargetDescription, TargetsMetadata, TimestampMetadata, VirtualTargetPath,
+    Delegations, Metadata, MetadataPath, RawSignedMetadata, Role, RootMetadata, SignedMetadata,
+    SnapshotMetadata, TargetDescription, TargetsMetadata, TimestampMetadata, VirtualTargetPath,
 };
+use crate::verify::{self, Verified};
 use crate::Result;
 
 /// Contains trusted TUF metadata and can be used to verify other metadata and targets.
 #[derive(Debug)]
 pub struct Tuf<D: DataInterchange> {
-    trusted_root: RootMetadata,
-    trusted_snapshot: Option<SnapshotMetadata>,
-    trusted_targets: Option<TargetsMetadata>,
-    trusted_timestamp: Option<TimestampMetadata>,
-    trusted_delegations: HashMap<MetadataPath, TargetsMetadata>,
+    trusted_root: Verified<RootMetadata>,
+    trusted_snapshot: Option<Verified<SnapshotMetadata>>,
+    trusted_targets: Option<Verified<TargetsMetadata>>,
+    trusted_timestamp: Option<Verified<TimestampMetadata>>,
+    trusted_delegations: HashMap<MetadataPath, Verified<TargetsMetadata>>,
     interchange: PhantomData<D>,
 }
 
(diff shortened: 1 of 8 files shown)


src/verify.rs
```diff
@@ -0,0 +1,160 @@
...
+        )));
+    }
+
+    // Everything looks good so deserialize the metadata.
+    //
+    // Note: Canonicalization (or any other transformation of data) could modify or filter out
+    // information about the data. Therefore, while we've confirmed the canonical bytes are signed,
+    // we shouldn't interpret this as if the raw bytes were signed. So we deserialize from the
+    // `canonical_bytes`, rather than from `raw_meta.as_bytes()`.
+    let verified_metadata = D::from_slice(&canonical_bytes)?;
```
wellsie1116, line 154: Instead of parsing 1 slice of raw bytes, this now parses 2 potentially different slices of raw bytes.

I would expect parsing the canonical bytes to be more expensive than parsing/typechecking from SignedMetadata::signed (D::RawData being, for example, serde_json::Value).   Why that functionality change here?

src/tuf.rs
```diff
@@ -30,30 +31,63 @@ impl<D: DataInterchange> Tuf<D> {
...
     /// the related method [`Tuf::from_root_with_trusted_keys`].
     pub fn from_trusted_root(signed_root: SignedMetadata<D, RootMetadata>) -> Result<Self> {
-        let verified = {
+        let verified_root = {
             let root = signed_root.assume_valid()?;
 
-            signed_root.verify(root.root().threshold(), root.root_keys())?
+            // Make sure the root signed itself.
+            let verified_root = verify::verify_signatures(
+                &signed_root.to_raw()?,
```
wellsie1116: to_raw was never intended to be used outside of generating metadata.  Why serialize the metadata just to re-parse it?

wellsie1116 (changes requested):
This CL says the goal is to be more strict on when metadata is deserialized, but it actually ends up parsing more unique metadata.  Per the TUF spec, signatures are on the canonical bytes of metadata, so it is unavoidable that rust-tuf will need to generically parse metadata (as DataInterchange::RawData) to generate the canonical byte ordering to verify signatures.  Maybe I'm missing something here, but I don't see the value in moving metadata parsing into its own module or the Verified wrapper.

I do like the idea of clearly encoding the TUF workflow in tuf.rs though, so the extra comments and moving the RawSignedMetadata to SignedMetadata conversions into tuf.rs are all nice to have.

src/tuf.rs
erickt: In the latest patch, I rewrote `Tuf::from_trusted_root` to take a `RawSignedMetadata`. That way the `Tuf` client can control all avenues towards deserializing then verifying metadata.

src/verify.rs
erickt: As we talked about offline, it is more expensive, but what we're trying to avoid is assuming just because we've verified the signatures of the canonical bytes, we can't assume that the `SignedMetadata` is signed. It's possible that canonicalization could change the metadata in some critical manner, such as removing a field. It's possible that the presence of that field later on could compromise the security of rust-tuf. So it's best if we throw away the raw metadata we used to extract out the signatures.

wellsie1116: Thanks for clarifying, I hadn't considered that D::RawData may strip out data when parsing generically parsing metadata.  serde_json::Value may not, but DataInterchange may be implemented outside this crate, so we shouldn't be assuming it works as intended.

wellsie1116 (approved):

PR #298, opened 2020-09-22
Make RepositoryStorage object-safe

erickt:
This changes the `RepositoryStorage` type to be object-safe by passing in the `AsyncRead` argument as a `&mut dyn AsyncRead`. Furthermore, Rust doesn't allow for a type like `Box<dyn A + B>`. So in order to make the `Repository` struct take a dyn boxed repository, this patch also adds a trait `RepositoryStorageProvider`, which has as a super trait `RepositoryStorage` and `RepositoryProvider`.

Test: Updated all the tests to work with this, and added a test to make sure these traits stay object-safe.

### DIFF
--- a/src/repository.rs
+++ b/src/repository.rs
@@ -68,24 +68,28 @@ where
     /// [`D::extension()`][extension], overwriting any existing metadata at that location.
     ///
     /// [extension]: crate::interchange::DataInterchange::extension
-    fn store_metadata<'a, R>(
+    fn store_metadata<'a>(
         &'a self,
         meta_path: &'a MetadataPath,
         version: &'a MetadataVersion,
-        metadata: R,
-    ) -> BoxFuture<'a, Result<()>>
-    where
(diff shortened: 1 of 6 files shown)


src/repository.rs
```diff
@@ -117,30 +121,85 @@ where
...
+    fn fetch_target<'a>(
+        &'a self,
+        target_path: &'a TargetPath,
+        target_description: &'a TargetDescription,
+    ) -> BoxFuture<'a, Result<Box<dyn AsyncRead + Send + Unpin>>> {
+        (**self).fetch_target(target_path, target_description)
+    }
+}
+
+impl<T, D> RepositoryStorageProvider<D> for Box<T>
```
wellsie1116: I wonder if we could define a blanket impl for all T: Storage + Provider instead of doing manual impls for each concrete type

wellsie1116 (approved):

src/repository.rs
erickt: done

Issue #292, opened 2020-09-17
Client::update_root should update from N+1, not fetch the latest root metadata


erickt:
The TUF spec states that the root metadata should be updated by continuing to try to update to version N+1 metadata until we get a not-found error. However, rust-tuf doesn't follow this scheme. Instead, `Client::update_root` fetches the latest root metadata, then sequentially updates from the current root version through the latest version. Unfortunately, this exposes the client to parsing untrusted TUF metadata. It's theoretically possible a parser exploit that could compromise a rust-tuf `Client` (see #291).

We should change `Client::update_root` to follow the spec to avoid this issue.

PR #300, opened 2020-10-08
Persist unversioned metadata to local store

erickt:
The TUF-1.0.9 spec states that the metadata should be written to non-volatile storage as an unversioned path after verification.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -463,6 +463,25 @@ where
     async fn update_root(&mut self) -> Result<bool> {
         let root_path = MetadataPath::from_role(&Role::Root);
 
+        // We don't follow the TUF-1.0.9 §5.1 on how to update the root metadata. It states:
+        //
+        // TUF-1.0.9 §5.1.2:
+        //
+        //     Try downloading version N+1 of the root metadata file, up to some W number of
+        //     bytes (because the size is unknown). The value for W is set by the authors of
+        //     the application using TUF. For example, W may be tens of kilobytes. The filename
+        //     used to download the root metadata file is of the fixed form
+        //     VERSION_NUMBER.FILENAME.EXT (e.g., 42.root.json). If this file is not available,
+        //     or we have downloaded more than Y number of root metadata files (because the
+        //     exact number is as yet unknown), then go to step 5.1.9. The value for Y is set
+        //     by the authors of the application using TUF. For example, Y may be 2^10.
+        //
+        // Instead, we fetch the latest available metadata (lets call the current version N and the
+        // latest version N+M), then we re-fetch all the metadata in betwee N and N+M.
+        //
+        // FIXME(#292): Consider rewriting this logic to follow the spec. By following the spec, we
+        // avoid the issue of having to use metadata (in order to extract the metadata version
+        // number) before we've verified it was signed correctly.
         let raw_latest_root = self
             .remote
             .fetch_metadata(
@@ -477,11 +496,7 @@ where
         // signed by the previous root metadata, which we can't check without knowing what version
         // this root metadata claims to be.
         let latest_version = {
-            // **WARNING**: By deserializing the metadata before verification, we are exposing us
-            // to parser exploits.
-            //
-            // FIXME(#292): Consider rewriting this logic to just fetching root version N+1 until
-            // we get a "Not Found" error. That allows us to avoid parsing the metadata here.
+            // FIXME(#292): See the note above.
             let latest_root = raw_latest_root.parse_untrusted()?;
             latest_root.parse_version_untrusted()?
         };
@@ -502,20 +517,7 @@ where
         for i in (self.tuf.trusted_root().version() + 1)..latest_version {
             let version = MetadataVersion::Number(i);
 
-            /////////////////////////////////////////
-            // TUF-1.0.9 §5.1.2:
-            //
-            //     Try downloading version N+1 of the root metadata file, up to some W number of
-            //     bytes (because the size is unknown). The value for W is set by the authors of
-            //     the application using TUF. For example, W may be tens of kilobytes. The filename
-            //     used to download the root metadata file is of the fixed form
-            //     VERSION_NUMBER.FILENAME.EXT (e.g., 42.root.json). If this file is not available,
-            //     or we have downloaded more than Y number of root metadata files (because the
-            //     exact number is as yet unknown), then go to step 5.1.9. The value for Y is set
-            //     by the authors of the application using TUF. For example, Y may be 2^10.
-
-            // FIXME(#292): Consider rewriting this logic to just fetching root version N+1 until
-            // we get a "Not Found" error.
+            // FIXME(#292): See the note above.
             let raw_signed_root = self
                 .remote
                 .fetch_metadata(&root_path, &version, self.config.max_root_length, None)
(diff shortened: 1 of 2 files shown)


src/client.rs
```diff
@@ -378,10 +377,14 @@ where
         // Only store the metadata after we have validated it.
         if fetched {
             client
-                .store_metadata(&root_path, &root_version, &raw_root)
+                .store_metadata(&root_path, &MetadataVersion::None, &raw_root)
                 .await;
 
-            // FIXME: should we also store the root as `MetadataVersion::None`?
+            // FIXME: This isn't a part of the spec, but we also store the versioned metadata. This
+            // allows us to initialize a repository from the local emtadata.
```
heartsucker: nit: typo on `metadata`

erickt: Fixed, thanks.

src/client.rs
```diff
@@ -490,6 +493,18 @@ where
...
+            // TUF-1.0.9 §5.1.2:
+            //
+            //     Try downloading version N+1 of the root metadata file, up to some W number of
+            //     bytes (because the size is unknown). The value for W is set by the authors of
+            //     the application using TUF. For example, W may be tens of kilobytes. The filename
+            //     used to download the root metadata file is of the fixed form
+            //     VERSION_NUMBER.FILENAME.EXT (e.g., 42.root.json). If this file is not available,
+            //     or we have downloaded more than Y number of root metadata files (because the
+            //     exact number is as yet unknown), then go to step 5.1.9. The value for Y is set
+            //     by the authors of the application using TUF. For example, Y may be 2^10.
```
ComputerDruid: This comment doesn't match the implementation very well, since this implementation walks until the expected version rather than until 404. There's a note about that above, but it should probably be noted boldly right next to this comment, lest the reader be confused into thinking that this comment is describing the behavior of the current implementation.

src/client.rs
```diff
@@ -500,34 +515,89 @@ where
...
+        // TUF-1.0.9 §5.1.9:
+        //
+        //     Check for a freeze attack. The latest known time MUST be lower than the expiration
+        //     timestamp in the trusted root metadata file (version N). If the trusted root
+        //     metadata file has expired, abort the update cycle, report the potential freeze
+        //     attack. On the next update cycle, begin at step 5.0 and version N of the root
+        //     metadata file.
+
+        // FIXME: we should move this logic into TUF to make it easier to review we implement the
+        // TUF spec.
```
ComputerDruid: huh? you cited the relevant part of the spec right above this? What do you mean?

src/client.rs
```diff
@@ -378,10 +377,14 @@ where
         // Only store the metadata after we have validated it.
         if fetched {
             client
-                .store_metadata(&root_path, &root_version, &raw_root)
+                .store_metadata(&root_path, &MetadataVersion::None, &raw_root)
```
ComputerDruid: It seems like this could roll back the metadata version in the local store, if the keys we are initializing from correspond to an old version. Should we avoid overwriting that here?

src/client.rs
```diff
@@ -500,34 +515,89 @@ where
...
+            // TUF-1.0.9 §5.1.7:
+            //
+            //     Persist root metadata. The client MUST write the file to non-volatile storage as
+            //     FILENAME.EXT (e.g. root.json).
+
+            self.store_metadata(&root_path, &MetadataVersion::None, &raw_signed_root)
+                .await;
+
+            // FIXME: This isn't a part of the spec, but we also store the versioned metadata. This
+            // allows us to initialize a repository from the local metadata.
```
ComputerDruid: This comment is weird to have as a FIXME. If we're doing it for a reason, why would it need to be fixed?

Also, when you say "this allows us to..." can you say where that code is, to make it easier to understand what this is enabling?

src/client.rs
```diff
@@ -538,10 +608,19 @@ where
...
+        if self.tuf.update_timestamp(&raw_signed_timestamp)?.is_some() {
+            /////////////////////////////////////////
+            // TUF-1.0.9 §5.2.4:
+            //
+            //     Persist timestamp metadata. The client MUST write the file to non-volatile
+            //     storage as FILENAME.EXT (e.g. timestamp.json).
+
+            self.store_metadata(
+                &timestamp_path,
+                &MetadataVersion::None,
```
ComputerDruid, line 630: Can you explain why this change is safe? It seems like if we were using local metadata at all this change would break that logic.

src/client.rs
erickt: Sure, I can add a reference to https://github.com/heartsucker/rust-tuf/issues/292 here.

src/client.rs
erickt: I'll rephrase this comment. We're checking the timestamp/snapshot/targets metadata for expiration in `tuf::Tuf`, I'd eventually like to move the root metadata expiration check into `tuf::Tuf` in order to reduce code duplication, and make it easier to audit rust-tuf's implementation of the spec, but I didn't want to go that far in this PR.

src/client.rs
erickt: While technically it's a breaking change, it's safe because `tuf::Client` doesn't actually use the local cache of the timestamp metadata. In fact, the only time we actually use the local store is in [Client::with_trusted_root_keys](https://github.com/heartsucker/rust-tuf/blob/develop/src/client.rs#L356), when we fetch the initial trusted root. There's no other place that reads from the store, other than tests.

It's possible someone could be relying on the side effect of these files being written out to local storage, but I'm not aware of anyone else actually using rust-tuf, so this should be safe.

src/client.rs
erickt: I'll remove this change. I think we'd only run into this if we called `Client::with_trusted_root_keys` with `MetadataVersion::None`. We probably should make that impossible though. I don't think it's safe to assume we can always trust that the initial trusted keys have always signed the latest root metadata.

src/client.rs
erickt: Sure, I'll switch to calling this a "NOTE". I used "FIXME" since I wanted to eventually get this updated in the spec (see https://github.com/theupdateframework/specification/issues/108).

src/client.rs
ComputerDruid: nit: it'd be nice if instead of saying "we should consider doing X" it explains that we _don't follow the spec you snippet you just inlined_.

Right now it says:
```
// TUF-1.0.9 §5.1.2: try downloading version N+1 ...

// FIXME: Consider downloading version N+1...
```

If you're paying close attention you can tell that that means the whole TUF spec text snippet there doesn't describe what we're actually doing, but I think you'd be much better served by a comment that clearly says we're not following the spec. Something like:

```
// We don't actually follow the spec here; instead, we...
// FIXME: consider following the spec instead of that.
```

src/client.rs
```diff
@@ -823,6 +922,12 @@ where
                 .update_delegation(&targets_role, delegation.role(), &raw_signed_meta)
             {
                 Ok(_) => {
+                    /////////////////////////////////////////
+                    // TUF-1.0.9 §5.2.4:
+                    //
+                    //     Persist timestamp metadata. The client MUST write the file to
+                    //     non-volatile storage as FILENAME.EXT (e.g. timestamp.json).
+
```
ComputerDruid: This seems like the wrong snippet? This is for timestamp, while this code is for delegated roles

ComputerDruid (approved):
Just notes about the new comments; change otherwise looks good

src/client.rs
erickt: good catch, thanks.

PR #302, opened 2020-10-22
Change update_root to follow TUF-1.0.9 section 5.1

erickt:
This switches Client::update_root to follow TUF-1.0.9 section 5.1 to update the root metadata. It now will fetch root N+1 until it receives a not-found error.

Test: This behavior is checked by the current tests. We could add a new test that makes sure we don't fetch `root.json`, but I'm not sure we get much value from a test like that.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -463,67 +463,42 @@ where
     async fn update_root(&mut self) -> Result<bool> {
         let root_path = MetadataPath::from_role(&Role::Root);
 
-        // We don't follow the TUF-1.0.9 §5.1 on how to update the root metadata. It states:
-        //
-        // TUF-1.0.9 §5.1.2:
-        //
-        //     Try downloading version N+1 of the root metadata file, up to some W number of
-        //     bytes (because the size is unknown). The value for W is set by the authors of
-        //     the application using TUF. For example, W may be tens of kilobytes. The filename
-        //     used to download the root metadata file is of the fixed form
-        //     VERSION_NUMBER.FILENAME.EXT (e.g., 42.root.json). If this file is not available,
-        //     or we have downloaded more than Y number of root metadata files (because the
-        //     exact number is as yet unknown), then go to step 5.1.9. The value for Y is set
-        //     by the authors of the application using TUF. For example, Y may be 2^10.
-        //
-        // Instead, we fetch the latest available metadata (lets call the current version N and the
-        // latest version N+M), then we re-fetch all the metadata in betwee N and N+M.
-        //
-        // FIXME(#292): Consider rewriting this logic to follow the spec. By following the spec, we
-        // avoid the issue of having to use metadata (in order to extract the metadata version
-        // number) before we've verified it was signed correctly.
-        let raw_latest_root = self
-            .remote
-            .fetch_metadata(
-                &root_path,
-                &MetadataVersion::None,
-                self.config.max_root_length,
-                None,
-            )
-            .await?;
+        let mut updated = false;
 
-        // Root metadata is signed by its own keys, but we should only trust it if it is also
-        // signed by the previous root metadata, which we can't check without knowing what version
-        // this root metadata claims to be.
-        let latest_version = {
-            // FIXME(#292): See the note above.
-            let latest_root = raw_latest_root.parse_untrusted()?;
-            latest_root.parse_version_untrusted()?
-        };
+        loop {
+            /////////////////////////////////////////
+            // TUF-1.0.9 §5.1.2:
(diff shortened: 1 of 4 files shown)


src/client.rs
```diff
@@ -463,67 +463,42 @@ where
...
+            };
 
-            // FIXME(#292): See the note above.
-            let raw_signed_root = self
-                .remote
-                .fetch_metadata(&root_path, &version, self.config.max_root_length, None)
-                .await?;
+            updated = true;
 
             if !self.tuf.update_root(&raw_signed_root)? {
```
kellerb: can we change Tuf::update_root to require that the new root version == current root version +1?

src/client.rs
```diff
@@ -463,67 +463,42 @@ where
...
+        loop {
+            /////////////////////////////////////////
+            // TUF-1.0.9 §5.1.2:
+            //
+            //     Try downloading version N+1 of the root metadata file, up to some W number of
+            //     bytes (because the size is unknown). The value for W is set by the authors of
+            //     the application using TUF. For example, W may be tens of kilobytes. The filename
+            //     used to download the root metadata file is of the fixed form
+            //     VERSION_NUMBER.FILENAME.EXT (e.g., 42.root.json). If this file is not available,
+            //     or we have downloaded more than Y number of root metadata files (because the
```
kellerb, line 477: does this implementation mean rust-tuf is hardcoding Y to infinity?

kellerb (changes requested):
For testing, since a key part of TUF is using the correct sequence of operations when interacting with untrusted repos, it seems like it would be good to have e.g. a RepositoryProvider type that allows us to verify the sequence of operations performed on it by Client, which would then make adding a test for this change easy.

Maybe this would be easy to make by creating a new RepositoryProvider that forwards to a wrapped EphemeralRepository after recording the method call.

src/client.rs
erickt: No, `Y` in our case is `self.config.max_root_length` down on line 484. By default it's 1 MiB.

src/client.rs
erickt: Yeah this was weird. I fixed that in another patch I'm working on, I'll pull it back here.

erickt (commented):
You convinced me to add a helper repository that'll track changes to repository storage and providers.

Note: I removed the `versioned_init` tests, since that logic is now captured in the `root_chain_update` tests.

src/client.rs
kellerb: `Y` is the number of new root metadata files to try downloading, i think `self.config.max_root_length` is `W`

src/repository/track_repo.rs
```diff
@@ -0,0 +1,165 @@
...
+    {
+        Track::FetchFound {
+            path: meta_path.clone(),
+            version: version.clone(),
+            metadata: String::from_utf8(metadata.into()).unwrap(),
+        }
+    }
+}
+
+/// Helper Repository wrapper that tracks all the fetches and stores for testing purposes.
```
kellerb: "... tracks all the <ins>metadata</ins> fetches ..."

src/tuf.rs
```diff
@@ -190,14 +189,12 @@ impl<D: DataInterchange> Tuf<D> {
...
-                info!(
-                    "Attempted to update root to new metadata with the same version. \
-                     Refusing to update."
-                );
-                return Ok(false);
-            } else if new_root.version() < trusted_root.version() {
+            let next_root_version = trusted_root
+                .version()
+                .checked_add(1)
+                .expect("root version should be less than max u32");
```
kellerb: should we error instead of panic here?

semi-related, 2^32-1 doesn't seem very high for a process that could be automated

kellerb (approved):
approved pending comments

src/client.rs
erickt: Oops, good catch. I filed #306 and added a FIXME to track that.

src/repository/track_repo.rs
erickt: Fixed, thanks.

src/tuf.rs
erickt: Sure, I'll switch. We'd never trip over this in practice though because the new root would fail to deserialize a 2^32 value into a u32.

We could bump the version numbers up to a u64 are worthwhile, but consider the case where it only took 1 millisecond to create or fetch a root metadata. Trying to generate or fetch `(2^32-1) - N` would still take approximately 50 days. Repositories and clients should instead have separate monitoring in place to catch runaway processes like that.

PR #303, opened 2020-10-22
Extend tests to check all the tuf::Client constructors

erickt:
This makes sure that all the tuf::Client tests go through the `test_versioned_init`s.


erickt:
Ended up doing this in a separate patch.

PR #304, opened 2020-10-22
Client::update should fail if we can't write metadata

erickt:
Note: This is built upon #307, so that should be reviewed and landed first.

Previously, `Client::update` would warn, but ultimately ignore, if we failed to persist metadata to the local store. This doesn't play well with trying to use TUF offline.

Consider the case where a client is doing an update for some new targets file. If we do an update, but we run out of space before writing the new targets file. If we reboot the device and lose our connection to the repository, we can no longer validate the old timestamp file.

This patch changes the client to error when this happens. This would allow a user of `tuf::Client` to handle this error (like clearing some files from local storage before re-trying the update). If a user instead wants the old behavior, they could initialize a client with a local repository that only warns if an error occurs.

Test: This adds a test case where where the client will err out if we fail to write an update.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1103,10 +1103,8 @@ mod test {
     use super::*;
     use crate::crypto::{HashAlgorithm, KeyType, PrivateKey, SignatureScheme};
     use crate::interchange::Json;
-    use crate::metadata::{
-        MetadataPath, MetadataVersion, RootMetadataBuilder, SnapshotMetadataBuilder,
-        TargetsMetadataBuilder, TimestampMetadataBuilder,
-    };
+    use crate::metadata::{MetadataPath, MetadataVersion};
+    use crate::repo_builder::RepoBuilder;
     use crate::repository::{EphemeralRepository, Track, TrackRepository};
     use chrono::prelude::*;
     use futures_executor::block_on;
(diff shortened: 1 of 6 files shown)


src/repository/error_repo.rs
```diff
@@ -0,0 +1,83 @@
...
+
+impl<R> ErrorRepository<R> {
+    pub(crate) fn new(repo: R) -> Self {
+        Self {
+            repo,
+            fail_stores: Arc::new(Mutex::new(false)),
+        }
+    }
+
+    pub(crate) fn fail_stores(&self, fail_stores: bool) {
```
wellsie1116: this method name implies it will fail all stores, but the RepositoryStorage impl only fails metadata stores.  A more specific method name or an impl that also fails target stores would be more clear to me.

src/client.rs
```diff
@@ -509,12 +487,14 @@ where
             //     Persist root metadata. The client MUST write the file to non-volatile storage as
             //     FILENAME.EXT (e.g. root.json).
 
-            self.store_metadata(&root_path, &MetadataVersion::None, &raw_signed_root)
-                .await;
+            self.local
+                .store_metadata(&root_path, &MetadataVersion::None, &raw_signed_root)
+                .await?;
```
wellsie1116, line 492: self.tuf was already updated with the new root metadata, so if this fails, then the caller takes some action to let this succeed, then calls update again, this loop won't fetch the root metadata again (because self.tuf already has it), resulting in the metadata never being stored locally.

Unfortunately, this ordering of operations aligns with the spec (5.1.6 and 5.1.7).  Maybe Client/Tuf should be tracking both a "trusted" root metadata and a "persisted" root metadata to ensure transient repository store errors don't lose metadata updates and eventually get persisted on retry?

src/client.rs
```diff
@@ -578,12 +558,13 @@ where
             //     Persist timestamp metadata. The client MUST write the file to non-volatile
             //     storage as FILENAME.EXT (e.g. timestamp.json).
 
-            self.store_metadata(
-                &timestamp_path,
-                &MetadataVersion::None,
-                &raw_signed_timestamp,
-            )
-            .await;
+            self.local
```
wellsie1116, line 561: There are similar issues with the ordering of trusting/persisting timestamp, snapshot, and targets metadata.  Though, in these instances, the spec is not clear as to an ordering of trusting/persisting these metadata.

As-is has an interesting property that if a client's local repository is completely broken and always fails stores, pounding on update() will eventually fetch and trust the most recent targets metadata, allowing clients to eventually resolve targets (once update() returns `Ok(false)` instead of `Err(_)`).  Is that preferable to not updating the trusted metadata when storing the metadata fails (so update() continuously returns `Err(_)` and target resolution never unblocks)?

src/client.rs
```diff
@@ -1488,4 +1471,70 @@ mod test {
...
+                .targets_version(2)
+                .snapshot_version(2)
+                .timestamp_version(2)
+                .commit()
+                .await
+                .unwrap();
+
+            local.fail_stores(true);
+
+            // The second update should fail.
```
wellsie1116, line 1536: Should an immediate third/fourth/fifth/... update fail?  As-is, this would eventually succeed (though with false instead of the expected true result).

wellsie1116 (changes requested):

src/repository/error_repo.rs
erickt: I'll rename to `fail_metadata_stores`.

src/client.rs
erickt: That's a good catch. I filed #309 to extend the `FileSystemRepository` to cache metadata if a write fails, and to to try to re-write it again, either during an update, manually, or some other mechanism.

src/client.rs
erickt: Heh, that is an interesting property. I'm not sure if that's actually a problem though, or at least one we need to address for this PR. I'll bring it up with the TUF community (see https://github.com/theupdateframework/specification/issues/131) to see if they can find any problems with this behavior though.

src/client.rs
erickt: I'll add a test to demonstrate this behavior, with a link to https://github.com/theupdateframework/specification/issues/131 to explain what's going on.

wellsie1116 (approved):

PR #305, opened 2020-10-23
Remove MetadataVersion::Hash

### LINKED ISSUE #254: Remove MetadataVersion::Hash

erickt:
TUF no longer supports prefixing metadata paths by a hash, so this removes support for them.

Closes #254

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -874,8 +874,21 @@ where
                 Err(e) => return (delegation.terminating(), Err(e)),
             };
 
+            /////////////////////////////////////////
+            // TUF-1.0.9 §5.4:
+            //
+            //     Download the top-level targets metadata file, up to either the number of bytes
+            //     specified in the snapshot metadata file, or some Z number of bytes. The value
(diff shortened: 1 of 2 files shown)


src/client.rs
```diff
@@ -874,8 +874,21 @@ where
                 Err(e) => return (delegation.terminating(), Err(e)),
             };
 
+            /////////////////////////////////////////
+            // TUF-1.0.9 §5.4:
+            //
+            //     Download the top-level targets metadata file, up to either the number of bytes
+            //     specified in the snapshot metadata file, or some Z number of bytes. The value
```
wittrock, line 881: I wish the spec was more explicit about Z being a max length for the downloaded file, but that's a fix for elsewhere.

wittrock (approved):
LGTM

PR #310, opened 2020-10-29
Initialize clients with local metadata

### LINKED ISSUE #301: Client should initialize itself from the local metadata
While `Client` persists metadata to the local store, the only time it actually tries to use it is when we try to fetch the initial [trusted root metadata](https://github.com/heartsucker/rust-tuf/blob/develop/src/client.rs#L356). We should change the `Client` to initialize itself with the rest of the local metadata.

erickt:
This updates the `tuf::Client` constructors to load the initial metadata from the local repository. It does this by refactoring `Client` to support updating from arbitrary repositories, and optionally writing the metadata back to the local repository (since during initialization we don't want to write back what we just read). One thing to note is that all but the root metadata is stored unversioned, so during initialization we are treating things as if consistent_snapshots is false.

In order to do this, this also changes `tuf::Repository::store_{metadata,targets` to take `&self`.

Test: This adds a few tests to check this behavior.
* Check that using all client constructors will load metadata during initialization.
* Check that we ignore errors if local metadata is missing, expired, or malformed.
* Check that we won't try to load metadata from the local repository after client initialization.

Closes #301

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -452,39 +452,8 @@ where
 
     /// Returns `true` if an update occurred and `false` otherwise.
     async fn update_root(&mut self) -> Result<bool> {
-        let updated = Self::update_root_with_repos(
-            &self.config,
-            &mut self.tuf,
-            Some(&self.local),
-            &self.remote,
-        )
-        .await?;
-
-        /////////////////////////////////////////
-        // TUF-1.0.9 §5.1.9:
-        //
-        //     Check for a freeze attack. The latest known time MUST be lower than the expiration
-        //     timestamp in the trusted root metadata file (version N). If the trusted root
-        //     metadata file has expired, abort the update cycle, report the potential freeze
-        //     attack. On the next update cycle, begin at step 5.0 and version N of the root
-        //     metadata file.
-
-        // TODO: Consider moving the root metadata expiration check into `tuf::Tuf`, since that's
-        // where we check timestamp/snapshot/targets/delegations for expiration.
-        if self.tuf.trusted_root().expires() <= &Utc::now() {
-            error!("Root metadata expired, potential freeze attack");
-            return Err(Error::ExpiredMetadata(Role::Root));
-        }
-
-        /////////////////////////////////////////
-        // TUF-1.0.5 §5.1.10:
-        //
-        //     Set whether consistent snapshots are used as per the trusted root metadata file (see
-        //     Section 4.3).
-
-        // FIXME: validate we are properly setting the consistent snapshot.
-
-        Ok(updated)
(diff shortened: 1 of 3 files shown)


src/client.rs
```diff
@@ -389,15 +372,51 @@ where
...
+            let _ta = Self::update_targets_with_repos(&mut tuf, None, &local, false).await?;
+
+            Ok(())
+        }
+        .await;
+
+        match res {
+            Ok(()) => {}
+            Err(Error::NotFound) => {
+                // Ignore if the local repository has expired metadata.
```
kellerb: i think this comment is wrong

src/client.rs
```diff
@@ -439,6 +458,50 @@ where
...
+        //     Check for a freeze attack. The latest known time MUST be lower than the expiration
+        //     timestamp in the trusted root metadata file (version N). If the trusted root
+        //     metadata file has expired, abort the update cycle, report the potential freeze
+        //     attack. On the next update cycle, begin at step 5.0 and version N of the root
+        //     metadata file.
+
+        // TODO: Consider moving the root metadata expiration check into `tuf::Tuf`, since that's
+        // where we check timestamp/snapshot/targets/delegations for expiration.
+        if self.tuf.trusted_root().expires() <= &Utc::now() {
+            error!("Root metadata expired, potential freeze attack");
```
kellerb: not for this cl, but maybe it would be better to have a separate error variant specifically for this type of problem

src/client.rs
```diff
@@ -439,6 +458,50 @@ where
...
+        //
+        //     Check for a freeze attack. The latest known time MUST be lower than the expiration
+        //     timestamp in the trusted root metadata file (version N). If the trusted root
+        //     metadata file has expired, abort the update cycle, report the potential freeze
+        //     attack. On the next update cycle, begin at step 5.0 and version N of the root
+        //     metadata file.
+
+        // TODO: Consider moving the root metadata expiration check into `tuf::Tuf`, since that's
+        // where we check timestamp/snapshot/targets/delegations for expiration.
+        if self.tuf.trusted_root().expires() <= &Utc::now() {
```
kellerb: why does this need to be done here and in `update_root_with_repos`?

kellerb (approved):

src/client.rs
erickt: Fixed

src/client.rs
erickt: Oops, artifact from a rebase. I'll remove this from `update_root`.

src/client.rs
erickt: ack. @wellsie1116 has suggested returning a type that allows the caller to accept or reject expired metadata, which we could explore at some point in time, but we can do that later.

kellerb (approved):

PR #313, opened 2021-02-09
Add Error::BadHttpStatus to expose http status codes

erickt:
This switches the http repository from returning an `Opaque` error into returning BadHttpStatus, which allows clients to perform structured logging when a remote server returns a variety of HTTP status codes.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -31,7 +31,7 @@ pub enum Error {
     #[error("http: {0}")]
     Http(http::Error),
 
-    /// Unexpected error
+    /// Unexpected HTTP response status.
     #[error("error getting {uri}: request failed with status code {code}")]
     BadHttpStatus {
         /// HTTP status code.
--- a/src/error.rs
+++ b/src/error.rs
@@ -31,6 +31,16 @@ pub enum Error {
     #[error("http: {0}")]
     Http(http::Error),
 
+    /// Unexpected error
+    #[error("error getting {uri}: request failed with status code {code}")]
+    BadHttpStatus {
+        /// HTTP status code.
+        code: http::StatusCode,
+
+        /// URI Resource that resulted in the error.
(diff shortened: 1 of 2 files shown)


src/repository/http.rs
```diff
@@ -228,10 +228,10 @@ where
             if status == StatusCode::NOT_FOUND {
                 Err(Error::NotFound)
             } else {
-                Err(Error::Opaque(format!(
-                    "Error getting {:?}: {:?}",
-                    self.uri, resp
-                )))
+                Err(Error::BadHttpStatus {
+                    code: status,
+                    uri: self.uri.to_string(),
```
wellsie1116: this field's documentation indicates this is the URI that was being fetched, but `self.uri` is just the base URI.  The local var `uri` had the full URI being requested.

src/error.rs
```diff
@@ -31,6 +31,16 @@ pub enum Error {
     #[error("http: {0}")]
     Http(http::Error),
 
+    /// Unexpected error
```
wellsie1116: Unexpected HTTP response status. ?

wellsie1116 (approved):

src/repository/http.rs
```diff
@@ -228,10 +228,10 @@ where
...
-                    "Error getting {:?}: {:?}",
-                    self.uri, resp
-                )))
+                Err(Error::BadHttpStatus {
+                    code: status,
+                    uri: uri.to_string(),
+                })
             }
         } else {
             Ok(resp)
```
dhobsd, line 237: WRT the intent of this change, it's worth noting the `status.is_success` check accepts _any_ 200-level response, which is not correct per the expected semantics of the request. We should probably be checking specifically for `StatusCode::OK` and forwarding anything else here.

PR #314, opened 2021-02-09
HttpRepository::get only works with 200

erickt:
Some of the HTTP success status codes don't actually make sense with `HttpRepository`, like "204 No Content". As implemented, this would result in strange decoding errors further in the update workflow. Instead, this patch changes `HttpRepository::get` to only successfully return `Ok(resp)` if the status is 200.


dhobsd (approved):
LGTM; this change results in expected semantics for non-OK responses in this response space.

PR #316, opened 2021-05-07
Make PrivateKey a trait

sadmac7000:


src/crypto.rs
```diff
@@ -327,48 +327,36 @@ impl<'de> Deserialize<'de> for KeyType {
...
+    /// Generate Ed25519 key bytes in pkcs8 format.
+    pub fn keypair() -> Result<Vec<u8>> {
+        Ed25519KeyPair::generate_pkcs8(&SystemRandom::new())
+            .map(|bytes| bytes.as_ref().to_vec())
+            .map_err(|_| Error::Opaque("Failed to generate Ed25519 key".into()))
     }
 
     /// Create a new `PrivateKey` from an ed25519 keypair, a 64 byte slice, where the first 32
     /// bytes are the ed25519 seed, and the second 32 bytes are the public key.
-    pub fn from_ed25519(key: &[u8]) -> Result<Self> {
```
erickt: This should be public

src/crypto.rs
```diff
@@ -381,20 +369,152 @@ impl PrivateKey {
...
+            private.public_key().as_ref().to_vec(),
+        )?;
+
+        Ok(Ed25519PrivateKey { private, public })
+    }
+}
+
+impl PrivateKey for Ed25519PrivateKey {
+    fn sign(&self, msg: &[u8]) -> Result<Signature> {
+        match &self.public.scheme {
```
erickt: I don't think it's necessary to do this check. Instead, we'd only need to check this when we're constructing the `Ed25519PrivateKey` while passing in a public key, but I don't think we should have any codepaths that do that.

src/crypto.rs
```diff
@@ -381,20 +369,152 @@ impl PrivateKey {
...
+        match pk8.stdin {
+            Some(ref mut stdin) => stdin.write_all(&gen.stdout)?,
+            None => return Err(Error::Opaque("openssl has no stdin".into())),
+        };
+
+        Ok(pk8.wait_with_output()?.stdout)
+    }
+
+    fn from_pkcs8(der_key: &[u8], scheme: SignatureScheme) -> Result<Self> {
+        if let SignatureScheme::Ed25519 = scheme {
```
erickt: to hedge against future changes, instead I'd recommend:

```
match scheme {
    SignatureScheme:: RsaSsaPssSha256 | SignatureScheme:: RsaSsaPssSha512 => {}
    _ => { return Err(Error::IllegalArgument(format!("RSA keys do not support the signing scheme {}", scheme))); }
}
```

Or as an alternative, we could change `SigningScheme` to be:

```
#[non_exhaustive]
enum SigningScheme {
    Ed25519,
    Rsa(RsaSigningScheme),
    ...
}

#[non_exhaustive]
enum RsaSigningScheme {
    RsaSsaPssSha256,
    RsaSsaPssSha512,
}
```

Then we could change this function to take an `RsaSigningScheme` and never have to worry about the wrong scheme being passed in. The only catch is we'd have to make sure that serializing/deserializing a `SigningScheme` should produce a flat string.

src/crypto.rs
```diff
@@ -381,20 +369,152 @@ impl PrivateKey {
...
-        Ok(PrivateKey { private, public })
+        Ok(RsaPrivateKey { private, public })
     }
+}
+
+impl PrivateKey for RsaPrivateKey {
+    fn sign(&self, msg: &[u8]) -> Result<Signature> {
+        let rng = SystemRandom::new();
+        let mut buf = vec![0; self.private.public_modulus_len()];
+        let scheme = match &self.public.scheme {
```
erickt, line 534: Same here, we shouldn't need to check the scheme here.

src/crypto.rs
```diff
@@ -381,20 +369,152 @@ impl PrivateKey {
...
+        })
+    }
+
+    fn public(&self) -> &PublicKey {
+        &self.public
+    }
+}
 
+/// Union of the several types of supported private key.
+pub enum AnyPrivateKey {
```
erickt: This should be `#[non_exhaustive]` so we could add more to it over time.

However, do we really need this functionality? It'd let us generically create keys from a PKCS#8 file, but all we're doing is trying to parse a file one way, then failing back to another. It doesn't look like any of our tests use this, so we could add it later if we need it.

src/crypto.rs
```diff
@@ -450,134 +570,13 @@ impl PrivateKey {
...
-            .map_err(|_| Error::Encoding("Could not parse key as PKCS#8v2".into()))?;
-
-        if key.public_modulus_len() < 256 {
-            return Err(Error::IllegalArgument(format!(
-                "RSA public modulus must be 2048 or greater. Found {}",
-                key.public_modulus_len() * 8
-            )));
+    /// Unwrap the specific private key contained in this object and return it as a boxed trait
+    /// object.
+    pub fn into_boxed(self) -> Box<dyn PrivateKey> {
```
erickt: We can leave the boxing up to the callers by implementing `PrivateKey` on this type, with:

```
impl PrivateKey for AnyPrivateKey {
    fn sign(&self, bytes: &[u8]) -> ... {
         match self {
             AnyPrivateKey::Ed25519(key) => key.sign(bytes),
             AnyPrivateKey::RSA(key) => key.sign(bytes)
     }
     ...
}
```

src/crypto.rs
sadmac7000: Done

src/crypto.rs
sadmac7000: Moved into a debug_assert

src/crypto.rs
sadmac7000: Did the simpler option.

src/crypto.rs
sadmac7000: Removed.

src/crypto.rs
sadmac7000: Done.

src/crypto.rs
sadmac7000: We need the match anyway for converting the RSA scheme type. Made the misc. branch an unreachable!()

PR #319, opened 2021-07-27
Implement Repository{Storage,Provider} for Arc<T>

erickt:


wellsie1116 (approved):
This has me wondering if we should try for a blanket impl of types that deref to a type that impls these traits, but I'm not sure how feasible/desirable that is.

PR #321, opened 2021-09-28
feat: Update hyper to 0.14

sunng87:


erickt:
Thanks for this! It's going to take me a little time to land this though, because Fuchsia is still on hyper 0.13.x, and it's taking us a while to update to 0.14.x. However, I'm planning on factoring out the hyper code into another crate, which will make it easier to target hyper versions.

PR #323, opened 2021-10-08
Add Ephemeral::{metadata, targets}

erickt:
This changes Ephemeral repository to allow access to the underlying metadata
and targets that were written to the repository. This allows for direct
modification, in case users want to directly add metadata to the repository.
This should be safe, because the repositories are considered untrusted, and
still needs to go through verification through the workflow before it is used
by the client.

### DIFF
--- a/src/repository/ephemeral.rs
+++ b/src/repository/ephemeral.rs
@@ -37,6 +37,16 @@ where
             _interchange: PhantomData,
         }
     }
+
+    /// Return a map for all the metadata in the repository.
+    pub fn metadata(&self) -> ArcHashMap<(MetadataPath, MetadataVersion), Arc<[u8]>> {
+        Arc::clone(&self.metadata)
+    }
+
+    /// Return a map for all the targets in the repository.
+    pub fn targets(&self) -> ArcHashMap<TargetPath, Arc<[u8]>> {
+        Arc::clone(&self.targets)
+    }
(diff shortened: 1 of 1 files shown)


src/repository/ephemeral.rs
```diff
@@ -37,6 +37,16 @@ where
             _interchange: PhantomData,
         }
     }
+
+    /// Return a map for all the metadata in the repository.
+    pub fn metadata(&self) -> ArcHashMap<(MetadataPath, MetadataVersion), Arc<[u8]>> {
```
wellsie1116, line 42: Will rustdoc make one of those unclickable links if ArcHashMap is not pub?  If so, consider exporting that type alias or expanding the alias here.

erickt: I just checked, and rustdoc will automatically expand these type aliases to `Arc<RwLock<HashMap<...>>>`.

wellsie1116 (approved):
Agreed this seems safe, it'd be possible to do this kind of thing already with a tempdir and FileSystemRepository.

I think a builder and/or more scoped CRUD APIs would be a bit nicer since it wouldn't need to expose data structure details of EphemeralRepository in the public interface, but it probably isn't worth the extra plumbing.

erickt:
Ended up doing this in a separate CL.

PR #324, opened 2021-10-08
Bump version to 0.3.0-beta1

erickt:
rust-tuf has been pretty stable for some time, so I'm planning to cut a
release for HEAD. I'm keeping this as beta1 because before release I
want to factor out hyper into a subcrate before release in order to
allow users to select hyper 0.13.x or 0.14.x.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "tuf"
 edition = "2018"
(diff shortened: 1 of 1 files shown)


wellsie1116 (approved):
Ah, nice.  Now it matches the README ;)
```
Warning: Beta Software
```

PR #331, opened 2021-12-16
Remove tuf::client::PathTranslator

erickt:
This removes tuf::client::PathTranslator. I'm not sure if this is a holdover from pre TUF-1.0 era, but as of right now I'm not aware of any strong use cases for translating paths. The only one I can come up with is if some repository wants to escape OS-specific characters, but that could be done in a repository wrapper, or in the FileSystemRepository itself.

Note that this was written on top of #334, so we should land that first to avoid excess review churn.

### DIFF
--- a/tuf/src/metadata.rs
+++ b/tuf/src/metadata.rs
@@ -1368,33 +1368,33 @@ impl<'de> Deserialize<'de> for SnapshotMetadata {
 
 /// Wrapper for the virtual path to a target.
 #[derive(Debug, Clone, PartialEq, Hash, Eq, PartialOrd, Ord, Serialize)]
-pub struct VirtualTargetPath(String);
+pub struct TargetPath(String);
 
-impl VirtualTargetPath {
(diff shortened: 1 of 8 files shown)


tuf/src/metadata.rs
```diff
@@ -1457,69 +1457,40 @@ impl VirtualTargetPath {
         self.matches_chain(&*new)
     }
 
+    /// Prefix the target path with a hash value to support TUF spec 5.5.2.
+    pub fn with_hash_prefix(&self, hash: &HashValue) -> Result<TargetPath> {
+        let mut components = self.components();
+
+        // The unwrap here is safe because we checked in `safe_path` that the path is not empty.
```
wittrock: Though why not just return an error if the pop fails? A little belt-and suspenders doesn't hurt here.

wittrock (approved):
LGTM, one comment about some copied code.

tuf/src/metadata.rs
erickt: good idea, added.

PR #333, opened 2021-12-16
Put RSA behind a feature flag

erickt:
RSA key support doesn't conform to the spec yet, so put it behind an `unstable_rsa` feature flag. This will allow us to eventually conform to the spec without making a breaking API change.

In addition, this marks all the key enums as #[non_exhaustive], so we can add new keys in the future without breaking our public APIs.

Note that this is built upon #334, #332, and #335. Those should be reviewed first to avoid review churn.


wittrock (approved):
LGTM

PR #335, opened 2021-12-17
Add test to verify generated metadata doesn't change

erickt:
The saved interop metadata diverged from what rust-tuf produced because:

* We no longer use keyid_hash_algorithms by default
* We changed writing metadata to just use `DataInterchange::canonicalize`, instead of `DataInterchange::to_writer`.

In order to avoid breakage, this PR:

* Adds a test to make sure that rust-tuf produces what is saved.
* Removes `DataInterchange::to_writer`, to avoid the risk of interacting with non-canonical data.
* Makes sure that signatures are sorted
* Changes `JsonPretty::canonicalize` to first use `Json::canonicalize`
* Regenerates the saved metadata.

Note that this patch was written on top of #334 and #331, so we should review those first to avoid review churn.


interop-tests/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+    private: String,
+}
+
+#[derive(Clone, Deserialize)]
+struct TestKeyPair {
+    #[serde(rename = "keytype")]
+    _keytype: KeyType,
+    #[serde(rename = "scheme")]
+    _scheme: SignatureScheme,
+    #[serde(rename = "keyid_hash_algorithms")]
```
wittrock: Do we still need this for backwards compat?

interop-tests/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+    let repo = FileSystemRepositoryBuilder::new(dir0)
+        .metadata_prefix(Path::new("repository"))
+        .targets_prefix(Path::new("repository").join("targets"))
+        .build()?;
+
+    update_root(&repo, &keys, None, 1, consistent_snapshot).await;
+    add_target(&repo, &keys, 0, consistent_snapshot).await;
+
+    let mut i: u8 = 1;
+    let rotations = vec![
```
wittrock, line 280: A comment here would be helpful

interop-tests/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+
+    let root_path = MetadataPath::from_role(&Role::Root);
+    repo.store_metadata(
+        &root_path,
+        &MetadataVersion::Number(version),
+        &mut root.to_raw().unwrap().as_bytes(),
+    )
+    .await
+    .unwrap();
+    repo.store_metadata(
```
wittrock, line 147: I know why you have to do this twice (to get both a versioned and unversioned root metadata), but a comment would be helpful as well, since it looks weird in review.

tuf/src/interchange/cjson/pretty.rs
```diff
@@ -1,15 +1,16 @@
...
 
 use super::Json;
 use crate::interchange::DataInterchange;
 use crate::Result;
 
 /// Pretty JSON data interchange.
 ///
-/// This is identical to [Json] in all manners except for the `to_writer` method. Instead of
-/// writing the metadata in the canonical format, it instead pretty prints the metadata.
+/// This is identical to [Json] in all manners except for the `canonicalize` method. Instead of
```
wittrock, line 11: It's possible that this change breaks clients who are relying on to_writer. I'm not sure what rust-tuf's backwards compat guarantees are, but should we create another type instead?

tuf/src/metadata.rs
```diff
@@ -435,6 +435,7 @@ where
         self.signatures
             .retain(|s| s.key_id() != private_key.public().key_id());
         self.signatures.push(sig);
+        self.signatures.sort();
```
wittrock, line 438: Were we just getting lucky with this before?

wittrock (commented):
Generally looking good, a few questions.

tuf/src/metadata.rs
erickt: Not necessarily. The order of signatures was entirely dependent on the order you called `SignedMetadata::add_signature`, so things were consistent if we stored the keys in an ordered list, which we've done in the tests. This change just makes sure clients don't run into non-deterministic output if they happen to sign things out of order.

tuf/src/interchange/cjson/pretty.rs
erickt: This is breaking compatibility, but we don't have any guarantees at the moment since we haven't released 0.3.0 yet. We shouldn't have trouble in fuchsia since we've vendored rust-tuf, and can upadte rust-tuf and any in-tree stuff in one CL.

 I'm mainly trying to do a lot of the cleanup like this so I can cut a 0.3.0 release.

interop-tests/src/lib.rs
erickt: Added some docs. Please let me know if that helped.

interop-tests/src/lib.rs
erickt: Nope, I removed it.

interop-tests/src/lib.rs
erickt: Added a comment.

interop-tests/src/lib.rs
wittrock: Yup, thanks.

wittrock (approved):

PR #336, opened 2021-12-17
Add a new repository builder

erickt:
This introduces a new repository builder. It greatly simplifies creating metadata for a repository. Here's a simple example that creates a repository, with a file, and commiting it to a repository.

```rust
let key = Ed25519PrivateKey::from_pkcs8("...").unwrap();
let mut repo = EphemeralRepository::new();
let _metadata = RepoBuilder::<Json>::create(&mut repo)
    .trusted_root_keys(&[&key])
    .trusted_targets_keys(&[&key])
    .trusted_snapshot_keys(&[&key])
    .trusted_timestamp_keys(&[&key])
    .add_target("foo/bar", Cursor::new(b"some file"))
    .commit()
    .await
    .unwrap();
```

It also supports generating new metadata derived off of a TUF database:

```rust
let mut client: tuf::Client = ...;
client.update().await().unwrap();

let _metadata2 = RepoBuilder::from_database(&mut repo, client.database())
    .trusted_root_key(&[&key])
    .trusted_targets_key(&[&key])
    .trusted_snapshot_key(&[&key])
    .trusted_timestamp_key(&[&key])
    .create_root()
    .unwrap()
    .commit()
    .await
    .unwrap();
```

Unfortunately RepoBuilder does not yet take care of generating target delegations. That will come in a future patch.

### DIFF
--- a/tuf/src/repo_builder.rs
+++ b/tuf/src/repo_builder.rs
@@ -1,223 +1,2099 @@
-use crate::{
-    crypto::{HashAlgorithm, PrivateKey},
-    interchange::DataInterchange,
-    metadata::{
-        Metadata, MetadataPath, MetadataVersion, RawSignedMetadata, Role, RootMetadata,
-        RootMetadataBuilder, SignedMetadataBuilder, SnapshotMetadata, SnapshotMetadataBuilder,
-        TargetsMetadata, TargetsMetadataBuilder, TimestampMetadata, TimestampMetadataBuilder,
+//! Repository Builder
+
+use {
+    crate::{
+        crypto::{self, HashAlgorithm, PrivateKey},
+        database::Database,
+        interchange::DataInterchange,
+        metadata::{
+            Metadata, MetadataDescription, MetadataPath, MetadataVersion, RawSignedMetadata,
+            RawSignedMetadataSet, RawSignedMetadataSetBuilder, Role, RootMetadata,
+            RootMetadataBuilder, SignedMetadataBuilder, SnapshotMetadata, SnapshotMetadataBuilder,
+            TargetDescription, TargetPath, TargetsMetadata, TargetsMetadataBuilder,
+            TimestampMetadata, TimestampMetadataBuilder,
+        },
+        repository::RepositoryStorage,
+        Error, Result,
     },
-    repository::{Repository, RepositoryStorage},
-    Result,
+    chrono::Utc,
+    futures_io::{AsyncRead, AsyncSeek},
+    futures_util::AsyncSeekExt as _,
+    std::{collections::HashMap, io::SeekFrom, marker::PhantomData},
 };
 
-// This helper builder simplifies the process of creating new metadata.
-//
-// FIXME: This is not ready yet for public use, it is only intended for internal testing until the
-// design is complete.
-pub(crate) struct RepoBuilder<'a, R, D>
-where
-    R: RepositoryStorage<D> + Sync,
-    D: DataInterchange + Sync,
-{
-    repo: Repository<R, D>,
-    root_keys: Vec<&'a dyn PrivateKey>,
-    targets_keys: Vec<&'a dyn PrivateKey>,
-    snapshot_keys: Vec<&'a dyn PrivateKey>,
-    timestamp_keys: Vec<&'a dyn PrivateKey>,
-    snapshot_version: u32,
-    timestamp_version: u32,
-    root_builder: RootMetadataBuilder,
-    targets_builder: Option<TargetsMetadataBuilder>,
+mod private {
+    use super::*;
+
+    pub trait Sealed {}
+
+    impl Sealed for Root {}
+    impl<D: DataInterchange> Sealed for Targets<D> {}
+    impl<D: DataInterchange> Sealed for Snapshot<D> {}
+    impl<D: DataInterchange> Sealed for Timestamp<D> {}
+    impl<D: DataInterchange> Sealed for Done<D> {}
 }
 
-impl<'a, R, D> RepoBuilder<'a, R, D>
-where
-    R: RepositoryStorage<D> + Sync,
-    D: DataInterchange + Sync,
-{
-    pub(crate) fn new(repo: R) -> Self {
-        let repo = Repository::new(repo);
+/// Trait to track each of the repository building states.
+pub trait State: private::Sealed {}
 
+/// State to create a root metadata.
+#[doc(hidden)]
+pub struct Root {
+    builder: RootMetadataBuilder,
+}
+
+/// State to create a targets metadata.
+#[doc(hidden)]
+pub struct Targets<D: DataInterchange> {
+    staged_root: Option<Staged<D, RootMetadata>>,
+    builder: TargetsMetadataBuilder,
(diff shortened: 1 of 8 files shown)


tuf/src/repo_builder.rs
```diff
@@ -1,223 +1,2099 @@
...
 
-// This helper builder simplifies the process of creating new metadata.
-//
-// FIXME: This is not ready yet for public use, it is only intended for internal testing until the
-// design is complete.
-pub(crate) struct RepoBuilder<'a, R, D>
+mod private {
+    use super::*;
+
+    pub trait Sealed {}
```
wittrock, line 30: A little strange to have a pub trait in a mod called `private`. Is this a common Rust idiom?

wittrock: Also, rustdoc pls :)

tuf/src/repo_builder.rs
```diff
@@ -1,223 +1,2099 @@
...
+
+/// The final state for building repository metadata.
+pub struct Done<D: DataInterchange> {
+    staged_root: Option<Staged<D, RootMetadata>>,
+    staged_targets: Option<Staged<D, TargetsMetadata>>,
+    staged_snapshot: Option<Staged<D, SnapshotMetadata>>,
+    staged_timestamp: Option<Staged<D, TimestampMetadata>>,
+}
+
+impl State for Root {}
```
wittrock: I think this should go nearer the definition of State

tuf/src/repo_builder.rs
```diff
@@ -1,223 +1,2099 @@
...
-            snapshot_keys: vec![],
-            timestamp_keys: vec![],
-            snapshot_version: 1,
-            timestamp_version: 1,
-            root_builder: RootMetadataBuilder::new(),
-            targets_builder: None,
+            ctx: RepoContext {
+                repo,
+                db: Some(db),
+                signing_root_keys: vec![],
```
wittrock, line 365: I am probably missing something, but shouldn't we initialize most of the context here? Seems weird to leave it blank.

tuf/src/repo_builder.rs
```diff
@@ -1,223 +1,2099 @@
...
         self
     }
 
-    pub(crate) fn with_root_builder<F>(mut self, f: F) -> Self
+    /// Create a root metadata using the default settings.
+    pub fn create_root(self) -> Result<RepoBuilder<'a, D, R, Targets<D>>> {
+        self.with_root_builder(|builder| builder)
+    }
+
+    /// Create a targets metadata using the default settings.
```
wittrock, line 464: Would be good to get more specific on what 'create' means here. It seems like it just directs the builder to do this _later_, rather than doing it now.

tuf/src/repo_builder.rs
```diff
@@ -1,223 +1,2099 @@
...
+        RepoBuilder {
+            ctx: self.ctx,
+            state: Targets::new(None),
+        }
+    }
+
+    /// Initialize a [RootMetadataBuilder] with the configured keys, and the
+    /// following defaults:
+    ///
+    /// * If the [RepoBuilder] was created for a new repository,  was created with a [Database], the version will be 1
```
wittrock: nit: line length is inconsistent.

tuf/src/repo_builder.rs
```diff
@@ -1,223 +1,2099 @@
...
+    ///
+    /// and pass it to the closure for further configuration. The builder then
+    /// will be used to generate a new root metadata and stage
+    ///
+    /// Generate a new metadata This function will generate a new root metadata,
```
wittrock: nit: missing period

tuf/src/repo_builder.rs
```diff
@@ -1,223 +1,2099 @@
...
+    /// * create a snapshot metadata if necessary.
+    pub fn create_timestamp(self) -> Result<RepoBuilder<'a, D, R, Done<D>>> {
+        self.create_targets_if_necessary()?.create_timestamp()
+    }
+
+    /// Create a new targets using the default settings if one is required.
+    /// For example:
+    ///
+    /// * This is new metadata.
+    /// * The metadata has expired.
```
wittrock, line 670: It seems like these are conditions that would require a new targets, but that's not clear from this doc comment.

wittrock (changes requested):
This looks quite good. Much simpler for clients. If you want to discuss the cost/benefits of requiring users to step through the typestate, happy to meet, but I like how clean it is now. 

Just a few questions and nits.

tuf/src/repo_builder.rs
erickt: I'll add some comments.

Yeah, this is implementing the [sealed trait pattern](https://rust-lang.github.io/api-guidelines/future-proofing.html). It's a way to declare a public trait, but make sure people outside the repository can't implement it.

tuf/src/repo_builder.rs
erickt: done

tuf/src/repo_builder.rs
erickt: This was one of the areas I was playing around with. What I'm trying to allow is that users should be able to generate metadata if they only have a subset of keys. One common workflow is that you want to automatically generate new timestamps to keep them from expiring. To do this, you typically want to leave the timestamp keys online. But generally you want to keep the root keys offline. This means that all these keys are ultimately optional.

So the two ways I approached this was here, and to also define a notion of a `KeySet`, which would then hold all these key values, and a `KeySetBuilder`, that allows you to make these functions. But I decided to squash the two together since I couldn't think of a logical reason why we'd want to work with the keyset as an independent thing. It'd just be something we'd make, then immediately pass to the builder.

I could go either way. Do you think the ux would be more obvious if you had to first gather up all the keys before trying to build a repo?

erickt: Also, I forgot to add a test to verify that the keys aren't required to be online. I added one.

tuf/src/repo_builder.rs
erickt: fixed

tuf/src/repo_builder.rs
erickt: fixed

tuf/src/repo_builder.rs
erickt: Would `stage_targets()` read better here and elsewhere? And maybe `stage_targets_if_necessary()`?

erickt: I renamed things from `create_*` and `with_*_builder` to to `stage_*` and `stage_*_with_builder`. I think that reads a bit better.

tuf/src/repo_builder.rs
erickt: I've rephrased it in the latest version. Does it read better?

erickt (commented):
I've pushed up a new version that tries to address all the comments. One of the major things I did was that I've gotten rid of most of the overloaded functions. It really cluttered up the rustdoc, and I think it'd be rather rare to actually use the overloads. We could always add them back if necessary.

tuf/src/repo_builder.rs
wittrock: Thanks!

wittrock (approved):
LGTM

PR #338, opened 2021-12-20
Clean up Repository{Storage,Provider} arguments

erickt:
This removes a few arguments from RepositoryProvider and RepositoryStorage which were never used, and reorders RepositoryStorage::store_target to match other arguments.

### DIFF
--- a/tuf/src/repository/ephemeral.rs
+++ b/tuf/src/repository/ephemeral.rs
@@ -8,10 +8,9 @@ use std::collections::HashMap;
 use std::marker::PhantomData;
 use std::sync::Arc;
 
-use crate::crypto::{HashAlgorithm, HashValue};
 use crate::error::Error;
 use crate::interchange::DataInterchange;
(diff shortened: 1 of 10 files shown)


tuf/src/repository/ephemeral.rs
```diff
@@ -147,20 +143,18 @@ mod test {
             let repo = EphemeralRepository::<Json>::new();
 
             let data: &[u8] = b"like tears in the rain";
-            let target_description =
```
wittrock, line 150: What was this meant to represent?

wittrock (approved):
LGTM.

tuf/src/repository/ephemeral.rs
erickt: `TargetDescription` contains things like the file hashes and any custom fields. A long time ago, we just had a single `Repository` trait, and implementations needed to do all the validation, but we eventually split the trait up into `RepositoryProvider` and `RepositoryStorage`, and added a separate internal `Repository` type that handles validation in one place. We just forgot to clean up these arguments when we made the split.

PR #339, opened 2021-12-22
Rename client fetching targets

erickt:
This patch:

* renames `Client::fetch_target` to `Client::fetch_target_to_local` to make it more obvious all it does is write the target to the local store.
* adds a new `Client::fetch_target`, which returns the target AsyncRead.
* removes `Client::fetch_target_to_writer`, since clients could implement that by using the result of `Client::fetch_target`.

### DIFF
--- a/tuf/src/client.rs
+++ b/tuf/src/client.rs
@@ -52,8 +52,7 @@
 //! ```
 
 use chrono::offset::Utc;
-use futures_io::{AsyncRead, AsyncWrite};
-use futures_util::io::copy;
+use futures_io::AsyncRead;
 use log::{error, warn};
 use std::collections::HashMap;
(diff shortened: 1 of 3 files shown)


tuf/src/client.rs
```diff
@@ -1016,6 +1000,30 @@ where
     }
 }
 
+// For some reason clippy thinks we can elide the lifetime, but rust needs it so
+// since the fetch future can hold a reference to repo items.
```
wittrock: nit: "so since the fetch future can hold"

wittrock (approved):
One nit, LGTM.

tuf/src/client.rs
erickt: fixed

PR #340, opened 2021-12-22
Remove synchronous io

erickt:
We had a few places where we performed synchronous io:

* DataInterchange::from_reader
* MetadataDescription::from_reader
* TargetDescription::from_reader
* TargetDescription::from_reader_with_custom

The first two are unnecessary since metadata is only working with `&[u8]`s. The last two we can switch to using asynchronous io.


j-wing:
LGTM.

PR #342, opened 2021-12-22
Allow client and db to trust metadata sets and dbs

erickt:
This adds the following methods:

* Client::from_database - establish trust from a database.
* Client::database - return a reference to the wrapped database.
* Database::from_trusted_metadata - Create a database that trusts the passed in metadata.
* Database::from_metadata_with_trusted_keys - Create a database that uses the metadata if the included root metadata was signed by the trusted keys.

### DIFF
--- a/tuf/src/client.rs
+++ b/tuf/src/client.rs
@@ -337,6 +337,17 @@ where
         Self::new(config, tuf, local, remote).await
     }
 
+    /// Create a new TUF client. It will trust and update the TUF database.
+    pub async fn from_database(
+        config: Config,
+        tuf: Database<D>,
+        local: L,
+        remote: R,
+    ) -> Result<Self> {
+        let (local, remote) = (Repository::new(local), Repository::new(remote));
+        Self::new(config, tuf, local, remote).await
+    }
+
     /// Construct a client with the given parts.
     ///
     /// Note: Since this was created by a prior [Client], it does not try to load
@@ -357,6 +368,7 @@ where
         }
     }
 
+    /// Create a new TUF client. It will trust this TUF database.
     async fn new(
         config: Config,
         mut tuf: Database<D>,
@@ -1265,6 +1277,7 @@ mod test {
         WithTrustedLocal,
         WithTrustedRoot,
         WithTrustedRootKeys,
+        FromDatabase,
     }
 
     #[test]
@@ -1356,6 +1369,13 @@ mod test {
         ))
     }
 
+    #[test]
+    fn from_database_loads_metadata_from_local_repo() {
+        block_on(constructors_load_metadata_from_local_repo(
+            ConstructorMode::FromDatabase,
+        ))
+    }
+
     async fn constructors_load_metadata_from_local_repo(constructor_mode: ConstructorMode) {
         // Store an expired root in the local store.
         let mut local = EphemeralRepository::<Json>::new();
@@ -1431,6 +1451,14 @@ mod test {
             )
             .await
             .unwrap(),
+            ConstructorMode::FromDatabase => Client::from_database(
+                Config::default(),
+                Database::from_trusted_root(&metadata1.root).unwrap(),
+                track_local,
+                track_remote,
+            )
+            .await
+            .unwrap(),
         };
 
         assert_eq!(client.tuf.trusted_root().version(), 1);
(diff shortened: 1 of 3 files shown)


tuf/src/client.rs
```diff
@@ -337,6 +337,17 @@ where
         Self::new(config, tuf, local, remote).await
     }
 
+    /// Create a new TUF client. It will trust and update TUF database.
```
wittrock: nit: and update _the_ TUF database

tuf/src/database.rs
```diff
@@ -99,21 +100,60 @@ impl<D: DataInterchange> Database<D> {
         })
     }
 
+    /// Create a new [`Database`] struct from a set of metadata that is assumed to be trusted. that
```
wittrock: "that is assumed to be trusted. that"

tuf/src/database.rs
```diff
@@ -99,21 +100,60 @@ impl<D: DataInterchange> Database<D> {
...
+    /// are used to verify the signed metadata. The signed root metadata must be signed with at
+    /// least a `root_threshold` of the provided root_keys. It is not necessary for the root
+    /// metadata to contain these keys.
```
wittrock: > It is not necessary for the root metadata to contain these keys.

Can you explain why that is? I'd think that the root metadata would have to contain the signing keys.

tuf/src/database.rs
```diff
@@ -99,21 +100,60 @@ impl<D: DataInterchange> Database<D> {
...
+        };
+
+        db.update_metadata_after_root(metadata)?;
+
+        Ok(db)
+    }
+
+    /// Create a new [`Database`] struct from a set of metadata that is assumed to be trusted.
+    ///
+    /// **WARNING**: This is trust-on-first-use (TOFU) and offers weaker security guarantees than
```
wittrock, line 130: Perhaps add a comment about when a developer _should_ use this function.

tuf/src/database.rs
```diff
@@ -99,21 +100,60 @@ impl<D: DataInterchange> Database<D> {
...
+    /// metadata to contain these keys.
+    pub fn from_metadata_with_trusted_keys<'a, I>(
+        metadata: &RawSignedMetadataSet<D>,
+        root_threshold: u32,
+        root_keys: I,
+    ) -> Result<Self>
+    where
+        I: IntoIterator<Item = &'a PublicKey>,
+    {
+        let mut db = if let Some(root) = metadata.root() {
```
wittrock: This reads to me as "make a database if there's root metadata, but don't use the root keys." Why would we want to override the signing keys but use the rest of the root metadata?

tuf/src/database.rs
```diff
@@ -152,6 +192,45 @@ impl<D: DataInterchange> Database<D> {
             .unwrap_or(0)
     }
 
+    /// Verify and update metadata.
```
wittrock: Add a note about what the return type means, which in this case is that we've updated some metadata.

tuf/src/database.rs
```diff
@@ -952,6 +1030,63 @@ mod test {
         );
     }
 
+    #[test]
+    fn from_trusted_metadata_success() {
```
wittrock, line 1042: Add a failure case here?

tuf/src/database.rs
```diff
@@ -1314,4 +1449,89 @@ mod test {
 
         assert!(tuf.update_targets(&raw_targets).is_err());
     }
+
+    #[test]
+    fn good_metadata_update() {
```
wittrock: This test name could be more descriptive. What's "good" about this metadata update?

wittrock (changes requested):

tuf/src/database.rs
erickt: In TUF, we only ever use keys from a trusted root metadata, but we need a way to bootstrap the problem. These keys in effect are acting like the trusted keys from an old trusted root metadata, and they go through the same verification process that we use when doing a root key rotation. We check that the old trusted root keys signed the new root metadata, but once we trust the new root metadata, we only will ever use the new root keys the next time we do an update.

The [spec](https://theupdateframework.github.io/specification/latest/#update-root) goes into a lot of detail for this.

tuf/src/database.rs
erickt: Added one, please let me know if it makes sense.

tuf/src/database.rs
erickt: Carrying over from the other thread, the `root_keys` are only used to establish trust with the root role in `metadata`. Once we've established trust (which happens when we call `Database::from_root_with_trusted_keys`), we'll only trust the keys in the now trusted root role, not the `root_keys`. Does that make sense?

tuf/src/database.rs
erickt: Done

tuf/src/database.rs
wittrock: Gotcha, thanks.

tuf/src/database.rs
erickt: done

tuf/src/database.rs
erickt: done

tuf/src/database.rs
wittrock: Yup, thanks.

tuf/src/database.rs
wittrock: Missed this one.

tuf/src/database.rs
wittrock: This is very helpful, thanks.

wittrock (approved):
Still one lingering comment in client.rs, but otherwise looks good.

PR #343, opened 2021-12-22
Optional metafile length and snapshot

erickt:
According to the [spec] it is optional for timestamp to contain the length and hash of the snapshot, and for snapshot to contain the length and hash for targets. This is considered safe because the metadata still needs to have it's metadata signature verified before used. However, the client must set an upper bound on the download size if the metadata does not specify the download length.

The upper bounds have been updated to sync with python-tuf's [defaults].

[spec]: https://theupdateframework.github.io/specification/latest/#metapath-hashes
[defaults]: https://github.com/theupdateframework/python-tuf/blob/30ba6e9f9ab25e0370e29ce574dada2d8809afa0/tuf/settings.py

### DIFF
--- a/tuf/src/client.rs
+++ b/tuf/src/client.rs
@@ -366,8 +366,10 @@ where
         let res = async {
             let _r = Self::update_root_with_repos(&config, &mut tuf, None, &local).await?;
             let _ts = Self::update_timestamp_with_repos(&config, &mut tuf, None, &local).await?;
-            let _sn = Self::update_snapshot_with_repos(&mut tuf, None, &local, false).await?;
-            let _ta = Self::update_targets_with_repos(&mut tuf, None, &local, false).await?;
+            let _sn =
+                Self::update_snapshot_with_repos(&config, &mut tuf, None, &local, false).await?;
+            let _ta =
+                Self::update_targets_with_repos(&config, &mut tuf, None, &local, false).await?;
 
             Ok(())
         }
@@ -643,6 +645,7 @@ where
     async fn update_snapshot(&mut self) -> Result<bool> {
         let consistent_snapshot = self.tuf.trusted_root().consistent_snapshot();
         Self::update_snapshot_with_repos(
+            &self.config,
             &mut self.tuf,
             Some(&mut self.local),
             &self.remote,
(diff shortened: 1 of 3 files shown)


tuf/src/client.rs
```diff
@@ -675,7 +679,16 @@ where
...
         }
 
-        let (alg, value) = crypto::hash_preference(snapshot_description.hashes())?;
+        let hash_data = {
+            let snapshot_hashes = snapshot_description.hashes();
+
+            if snapshot_hashes.is_empty() {
+                None
+            } else {
+                let (alg, value) = crypto::hash_preference(snapshot_description.hashes())?;
```
wittrock: I think this line should just use snapshot_hashes again.

tuf/src/client.rs
```diff
@@ -753,7 +763,16 @@ where
...
         }
 
-        let (alg, value) = crypto::hash_preference(targets_description.hashes())?;
+        let hash_data = {
+            let targets_hashes = targets_description.hashes();
+
+            if targets_hashes.is_empty() {
+                None
+            } else {
+                let (alg, value) = crypto::hash_preference(targets_description.hashes())?;
```
wittrock: Same comment as above, just use targets_hashes.

tuf/src/client.rs
```diff
@@ -2238,4 +2265,91 @@ mod test {
             );
         })
     }
+
+    #[test]
+    fn client_can_update_with_unknown_len_and_hashes() {
```
wittrock, line 2270: I may have missed it, but is there a test using defined lengths? If not, we should have one.

wittrock: Also, please fix the compiler/clippy errors below.

tuf/src/metadata.rs
```diff
@@ -2562,6 +2562,34 @@ mod test {
...
+    #[test]
+    fn serde_timestamp_metadata_optional_length_and_hashes() {
+        let description = MetadataDescription::new(1, None, HashMap::new()).unwrap();
+
+        let timestamp = TimestampMetadataBuilder::from_metadata_description(description)
+            .expires(Utc.ymd(2017, 1, 1).and_hms(0, 0, 0))
+            .build()
+            .unwrap();
+
+        let jsn = json!({
```
wittrock, line 2575: The name of this test implies to me that we're looking to serialize a length here, but it doesn't appear in this json. Perhaps modify the test name to explicitly say "without", so `serde_timestamp_metadata_without_length_and_hashes`

wittrock (changes requested):

wittrock (approved):

PR #344, opened 2021-12-23
Validate all hashes; try all target hash prefixes

erickt:
This changes how the client downloads metadata and targets. Previously we would only make one attempt to download the most preferred hash algorithm. However, according to [update-snapshot] and [update-targets], we should verify all hashes are valid. This is done by moving much of the hash logic into `tuf::repository::Repository`.

Furthermore, [fetch-target] says that when we are downloading a target with consistent snapshots are enabled, we will now attempt to download the target for each hash prefix.

[update-snapshot]: https://theupdateframework.github.io/specification/v1.0.26/#update-snapshot
[update-targets]: https://theupdateframework.github.io/specification/v1.0.26/#update-targets
[fetch-target]: https://theupdateframework.github.io/specification/v1.0.26/#fetch-target

### DIFF
--- a/tuf/src/client.rs
+++ b/tuf/src/client.rs
@@ -155,7 +155,7 @@ where
         let root_version = MetadataVersion::Number(1);
 
         let raw_root: RawSignedMetadata<_, RootMetadata> = local
-            .fetch_metadata(&root_path, &root_version, config.max_root_length, None)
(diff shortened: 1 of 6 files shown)


wittrock (approved):
LGTM

PR #346, opened 2022-01-07
Add ephemeral and file system repo transactions

erickt:
This introduces the notion of a repository transaction, and implements it for EphemeralRepository and FileSystemRepository. These are special repositories that are designed to be used when creating new TUF metadata. Rather than writing directly to the repository, these stage any queued up metadata in a staging location, and allows users to as-atomic-as-possible write that data back to the source repository when everything has been staged.

One caveat. While the FileSystemTransaction can atomically write each file, it may be possible for a system error to cause `commit()` to fail partway through writing files.

### DIFF
--- a/tuf/src/repository/ephemeral.rs
+++ b/tuf/src/repository/ephemeral.rs
@@ -33,11 +33,11 @@ where
         }
     }
 
-    /// Returns a [EphemeralTransaction] for manipulating this repository. This allows callers to
-    /// stage a number of mutations, and optionally commit them all at once.
-    pub fn transaction(&mut self) -> EphemeralTransaction<'_, D> {
-        EphemeralTransaction {
-            commit_repo: self,
+    /// Returns a [EphemeralBatchUpdate] for manipulating this repository. This allows callers to
+    /// stage a number of mutations, and optionally atomically write them all at once.
+    pub fn batch_update(&mut self) -> EphemeralBatchUpdate<'_, D> {
+        EphemeralBatchUpdate {
+            parent_repo: self,
             staging_repo: EphemeralRepository::new(),
         }
     }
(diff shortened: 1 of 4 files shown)


tuf/src/repository/ephemeral.rs
```diff
@@ -122,10 +121,104 @@ where
     }
 }
 
+/// [EphemeralTransaction] is a special repository that is designed to atomically commit metadata
+/// and targets to an [EphemeralRepository]. It can be used as a normal repository.
+///
+/// Note: `EphemeralTransaction::commit()` must be called in order to write the metadata and targets
+/// to the [EphemeralRepository]. Otherwise any stored file will be lost on drop.
```
wittrock: s/stored file/queued changes

tuf/src/repository/ephemeral.rs
```diff
@@ -122,10 +121,104 @@ where
     }
 }
 
+/// [EphemeralTransaction] is a special repository that is designed to atomically commit metadata
```
wittrock: If this isn't truly atomic (i.e. it either all succeeds or all fails), I would stay away from 'atomic'. Perhaps instead use language like `commit metadata in a batch`?

tuf/src/repository/ephemeral.rs
```diff
@@ -32,6 +32,15 @@ where
             _interchange: PhantomData,
         }
     }
+
+    /// Returns a [EphemeralTransaction] for manipulating this repository. This allows callers to
+    /// stage a number of mutations, and optionally commit them all at once.
+    pub fn transaction(&mut self) -> EphemeralTransaction<'_, D> {
```
wittrock: I'd name this something else if it doesn't have database-like transaction semantics, in particular that partially-applied changes are rolled back on failure. Perhaps `batch_update`?

tuf/src/repository/file_system.rs
```diff
@@ -107,6 +108,42 @@ where
             .targets_prefix("targets")
             .build()
     }
+
+    /// Returns a [FileSystemTransaction] for manipulating this repository. This allows callers to
+    /// stage a number of mutations, and optionally commit them all at once.
+    pub fn transaction(&mut self) -> FileSystemTransaction<D> {
```
wittrock: Same comment about naming here: unless we can guarantee transaction semantics, we should call this a `BatchUpdate` or something.

tuf/src/repository/file_system.rs
```diff
@@ -199,6 +216,117 @@ where
     }
 }
 
+/// [FileSystemTransaction] is a special repository that is designed to atomically commit metadata
```
wittrock: Same comment here about Transaction and atomic commit wording.

wittrock (changes requested):
Mostly comments about naming, but functionality looks fine.

tuf/src/repository/ephemeral.rs
erickt: `EphemeralTransaction::commit()` is actually atomic, since it cannot fail (other than if something panics, but that would tear down the process). It's `FileSystemTransaction::commit()` that's not truly atomic.

tuf/src/repository/file_system.rs
erickt: Fixed it up. Let me know how it reads.

tuf/src/repository/file_system.rs
erickt: Renamed to `FileSystemBatchUpdate`.

tuf/src/repository/ephemeral.rs
erickt: Renamed to `batch_update()`.

tuf/src/repository/ephemeral.rs
erickt: Even though it's atomic, I renamed it to EphemeralBatchUpdate to be consistent.

wittrock (approved):
LGTM, thanks!

PR #348, opened 2022-01-27
Integrate with coveralls

erickt:


wittrock (approved):
Thanks for implementing this so quickly!

PR #353, opened 2022-02-03
Clean up some client and RepoBuilder APIs

erickt:
This exposes the database, and repositories from the client, remotes some now redundant APIs, and renames RepoBuilder::with_timestamp_builder() to stage_timestamp_with_builder().

### DIFF
--- a/tuf/src/client.rs
+++ b/tuf/src/client.rs
@@ -54,7 +54,6 @@
 use chrono::offset::Utc;
 use futures_io::AsyncRead;
 use log::{error, warn};
-use std::collections::HashMap;
 use std::future::Future;
 use std::pin::Pin;
 
@@ -64,7 +63,7 @@ use crate::error::Error;
 use crate::interchange::DataInterchange;
 use crate::metadata::{
     Metadata, MetadataPath, MetadataVersion, RawSignedMetadata, Role, RootMetadata,
-    SnapshotMetadata, TargetDescription, TargetPath, TargetsMetadata, TimestampMetadata,
+    SnapshotMetadata, TargetDescription, TargetPath, TargetsMetadata,
 };
 use crate::repository::{Repository, RepositoryProvider, RepositoryStorage};
 use crate::verify::Verified;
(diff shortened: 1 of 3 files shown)


tuf/src/client.rs
```diff
@@ -430,54 +429,34 @@ where
         }
     }
 
-    /// Returns the current trusted root version.
-    pub fn root_version(&self) -> u32 {
-        self.tuf.trusted_root().version()
+    /// Returns a reference to the TUF database.
+    pub fn database(&self) -> &Database<D> {
```
wellsie1116, line 433: Is there a concern with giving client holders access to potentially read untrusted metadata, mutate the Database, and mutate/inspect local/remote repos?  Sure, some caller constructed a tuf::Client from L/R originally, but this opens up the holder of a tuf::Client to access potentially invalid target metadata.

erickt: Is your main concern `Client::database_mut()`? I could drop that since that's not something I really need at the moment.

That said, I'm not too worried about someone accessing that though for a few reasons:

* If you got a `&mut client`, you could always do `mem::swap(&mut client, &mut my_compromised_client)`.
* If you own the client, you could reform it with `Client::into_parts()` and `Client::from_parts(parts)`.
* Holders of `&mut client` could only inject signed metadata that meets the tuf schema. At that point, as far as the tuf spec is concerned the new metadata is valid. That could have been a complication if we tried to change the database during an update, but I've gotten rid of all the interior mutation, so we're protected from problems there by the borrow checker.

wellsie1116: I think of instances of tuf::Client as the capability to update metadata, lookup verified target descriptions, and read verified target data.  My main concern is that giving mutable field access can break any invariants we have between fields of Client or even within a field that wasn't intended to be publicly accessible.  For example, with `database_mut`, a user of Client could do a "partial update", calling only `update_targets` and never updating the root metadata, potentially trusting metadata that should be rejected by a key rotation that the Database never saw.  `Database` only implements the TUF spec when driven by `Client`.

I did not know of into_parts/from_parts.  Nevermind then, this doesn't expose any more mutable functionality than was already provided with Parts and mem::swap, and clients would have to go pretty far out of their way to fetch/parse unverified remote metadata through remote_repo()

wellsie1116 (approved):

PR #358, opened 2022-05-11
feat: add workflow responsible for notifying of new TUF spec releases

rdimitrov:
The following PR adds a workflow that keeps track and creates an issue for notifying of new TUF specification releases.

The version (v1.0.29) of the spec which is currently used for the verification is added without verifying if this is indeed the version that the `rust-tuf` project is compliant with. In that sense, please advise which version to use and I'll update the PR accordingly.

**Note:** Dependent on - https://github.com/theupdateframework/specification/pull/224

Signed-off-by: Radoslav Dimitrov <dimitrovr@vmware.com>

### DIFF
--- a/.github/workflows/specification-version-check.yml
+++ b/.github/workflows/specification-version-check.yml
@@ -1,0 +1,14 @@
+on:
+  schedule:
+    - cron: "0 13 * * *"
+  workflow_dispatch:
+name: Specification version check
+jobs:
+  # Get the latest TUF specification release and open an issue (if needed)
+  specification-bump-check:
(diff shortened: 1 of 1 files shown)


erickt:
Neat! I'll wait until this lands upstream then merge it in.

rdimitrov:
> Neat! I'll wait until this lands upstream then merge it in.

Great! I'll let you know when that happens 👍

rdimitrov:
@erickt - Hey, just so you know the dependent PR is now merged in the specification repo 👍

PR #360, opened 2022-05-31
Consolidate Database::update_{targets,delegated_targets}

erickt:
The TUF update flow for targets and delegated targets is mostly the same, just that the targets keys are in the root metadata, and the delegated target keys are from the delegating target metadata.

This is built upon #359, so that should be reviewed first. I'll rebase this on 359 once it lands.

### DIFF
--- a/tuf/src/database.rs
+++ b/tuf/src/database.rs
@@ -168,34 +168,6 @@ impl<D: DataInterchange> Database<D> {
         &self.trusted_delegations
     }
 
-    fn trusted_timestamp_version(&self) -> u32 {
-        self.trusted_timestamp
-            .as_ref()
-            .map(|t| t.version())
-            .unwrap_or(0)
-    }
-
-    fn trusted_snapshot_version(&self) -> u32 {
-        self.trusted_snapshot
-            .as_ref()
-            .map(|t| t.version())
-            .unwrap_or(0)
-    }
-
-    fn trusted_targets_version(&self) -> u32 {
-        self.trusted_targets
-            .as_ref()
-            .map(|t| t.version())
-            .unwrap_or(0)
-    }
-
-    fn trusted_delegation_version(&self, role: &MetadataPath) -> u32 {
-        self.trusted_delegations
-            .get(role)
-            .map(|t| t.version())
-            .unwrap_or(0)
-    }
-
     /// Verify and update metadata. Returns true if any of the metadata was updated.
     pub fn update_metadata(&mut self, metadata: &RawSignedMetadataSet<D>) -> Result<bool> {
         let updated = if let Some(root) = metadata.root() {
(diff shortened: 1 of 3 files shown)


tuf/src/database.rs
```diff
@@ -370,20 +342,20 @@ impl<D: DataInterchange> Database<D> {
...
+                match new_timestamp.version().cmp(&trusted_timestamp.version()) {
+                    Ordering::Less => {
+                        return Err(Error::VerificationFailure(format!(
+                            "Attempted to roll back timestamp metadata at version {} to {}.",
+                            trusted_timestamp.version(),
+                            new_timestamp.version()
+                        )));
+                    }
+                    Ordering::Equal => {
+                        return Ok(None);
```
kellerb, line 355: noticing that before the `trusted_timestamp_version` defaulted to `0`, so if the `new_timestamp` were also `0` we would take the early return here, but now if `self.trusted_timestamp` is `None` we skip this, which is like `Greater`

tuf/src/database.rs
```diff
@@ -435,24 +409,25 @@ impl<D: DataInterchange> Database<D> {
...
+                {
+                    Ordering::Less => {
+                        return Err(Error::VerificationFailure(format!(
+                            "Attempted to roll back snapshot metadata at version {} to {}.",
+                            trusted_snapshot.version(),
+                            trusted_timestamp.snapshot().version()
+                        )));
+                    }
+                    Ordering::Equal => {
+                        return Ok(false);
```
kellerb, line 427: ditto

kellerb (approved):

tuf/src/database.rs
erickt: The TUF spec actually forbids metadata having a version 0, so the metadata struct constructors error out if the version is ever 0:

* [root](https://github.com/theupdateframework/rust-tuf/blob/develop/tuf/src/metadata.rs#L775)
* [targets](https://github.com/theupdateframework/rust-tuf/blob/develop/tuf/src/metadata.rs#L1858)
* [snapshot](https://github.com/theupdateframework/rust-tuf/blob/develop/tuf/src/metadata.rs#L1432)
* [timestamp](https://github.com/theupdateframework/rust-tuf/blob/develop/tuf/src/metadata.rs#L1164)
* [metadata description in timestamp/snapshot](https://github.com/theupdateframework/rust-tuf/blob/develop/tuf/src/metadata.rs#L1164)

Previously we were being a bit clever by using an illegal value of `0` here instead of an option to represent a missing value. So in this case, we should only exit early here and in the other places I changed this pattern only if we have a trusted metadata. Otherwise we should always finish the upadte process.

It'd probably make sense to instead use https://doc.rust-lang.org/stable/std/num/struct.NonZeroU32.html to make sure this is semantically correct. We can do that later though.

tuf/src/database.rs
erickt: ditto from above.

PR #361, opened 2022-05-31
Implement update start time

erickt:
This patch implements [TUF 1.0.30 section 5.1.1], where we record the update start time at the beginning of the client update workflow, and compare all expirations to this. This allows us to avoid the risk of the metadata expiring during the update process.
    
In addition, this adds `Client::update_with_start_time()`, which allows the client to specify the start time to use, rather than the current time. This allows the client to work with expired metadata (and so there is a warning in the documentation), it could also be used to share expiration time with TUF and other external processes.
    
[TUF 1.0.30 section 5.1.1]: https://theupdateframework.github.io/specification/latest/#fix-time

### DIFF
--- a/tuf/src/client.rs
+++ b/tuf/src/client.rs
@@ -51,7 +51,7 @@
 //! # }
 //! ```
 
-use chrono::offset::Utc;
+use chrono::{offset::Utc, DateTime};
 use futures_io::AsyncRead;
 use log::{error, warn};
 use std::future::Future;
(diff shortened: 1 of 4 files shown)


tuf/src/client.rs
```diff
@@ -846,20 +900,41 @@ where
...
+    ) -> Result<TargetDescription> {
+        self.fetch_target_description_with_start_time(target, &Utc::now())
+            .await
+    }
+
+    /// Fetch a target description from the remote repo and return it.
+    pub async fn fetch_target_description_with_start_time(
+        &mut self,
+        target: &TargetPath,
+        now: &DateTime<Utc>,
```
kellerb: nit: the `now` parameter is called `start_time` in some of the other fns

kellerb (approved):

tuf/src/client.rs
erickt: ah you caught me. Fixed.

PR #363, opened 2022-07-14
Improve errors

erickt:
This PR cleans up the errors to include better information:

* missing metadata or target path is now included in relevant errors.
* unused error variants are removed.
* makes variants for common errors.

### DIFF
--- a/tuf/src/error.rs
+++ b/tuf/src/error.rs
@@ -2,10 +2,12 @@
 
 use data_encoding::DecodeError;
 use std::io;
-use std::path::Path;
 use thiserror::Error;
 
-use crate::metadata::MetadataPath;
+use crate::metadata::{MetadataPath, MetadataVersion, TargetPath};
+
+/// Alias for `Result<T, Error>`.
+pub type Result<T> = std::result::Result<T, Error>;
 
 /// Error type for all TUF related errors.
 #[non_exhaustive]
@@ -13,7 +15,7 @@ use crate::metadata::MetadataPath;
 pub enum Error {
     /// The metadata had a bad signature.
     #[error("bad signature")]
-    BadSignature,
+    BadSignature(MetadataPath),
 
     /// There was a problem encoding or decoding.
     #[error("encoding: {0}")]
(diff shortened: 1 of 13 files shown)


wellsie1116 (changes requested):
errors improved, nice.  Looks like there are some CI failures to address first.

tuf/src/error.rs
```diff
@@ -45,11 +47,35 @@ pub enum Error {
...
+    /// A tempfile persist error occurred.
+    #[error("tempfile: {0}")]
+    TempfilePersist(#[from] tempfile::PersistError),
+
+    /// A tempfile path persist error occurred.
+    #[error("tempfile: {0}")]
+    TempfilePathPersist(#[from] tempfile::PathPersistError),
+
+    /// A derp error occurred.
+    #[error("derp: {0}")]
```
wellsie1116: "derp" is a cute name, but maybe we should indicate what that means in the error string. "der parse: {0}"?

erickt:
Those pesky features. Should hopefully be fixed now.

tuf/src/error.rs
```diff
@@ -2,18 +2,20 @@
...
+
+/// Alias for `Result<T, Error>`.
+pub type Result<T> = std::result::Result<T, Error>;
 
 /// Error type for all TUF related errors.
 #[non_exhaustive]
 #[derive(Error, Debug)]
 pub enum Error {
     /// The metadata had a bad signature.
     #[error("bad signature")]
```
wellsie1116: let's include the path in the error string

wellsie1116 (commented):
TIL clicking the wrong start review button deletes all my comments

tuf/src/error.rs
```diff
@@ -41,11 +43,35 @@ pub enum Error {
         uri: String,
     },
 
+    /// An IO error occurred.
+    #[error("IO: {0}")]
```
kellerb: we've been not including source errors in the display text and then using anyhow to stringify the errors for logging, which itself includes the source errors in the stringification

not sure if that pattern would be useful here

tuf/src/metadata.rs
```diff
@@ -2550,11 +2569,17 @@ mod test {
             serde_json::from_value(jsn).unwrap();
         let raw_root = decoded.to_raw().unwrap();
         assert_matches!(
-            verify_signatures(&raw_root, 2, &[root_key.public().clone()]),
-            Err(Error::VerificationFailure(_))
+            verify_signatures(&MetadataPath::root(), &raw_root, 2, &[root_key.public().clone()]),
+            Err(Error::MetadataMissingSignatures { role, number_of_valid_signatures: 1, threshold: 2 })
```
kellerb: some line wraps in the following assert_matches

tuf/src/error.rs
erickt: done

tuf/src/metadata.rs
erickt: done

tuf/src/error.rs
erickt: I switched back to the old error writing the error in the `Encoding` variant. I think I want to eventually hide derp properly, but I figure I can do that later.

tuf/src/error.rs
erickt: seems reasonable. done.

kellerb (approved):

PR #366, opened 2022-07-14
Address clippy lints

erickt:


tuf/src/metadata.rs
```diff
@@ -3036,7 +3036,7 @@ mod test {
     //                              (,,(,    \.         `.   ____,-`.,
     //                           (,'     `/   \.   ,--.___`.'
     //                       ,  ,'  ,--.  `,   \.;'         `
-    //                        `{o, {    \  :    \;
+    //                        `(o, /    \  :    \;
```
kellerb, line 3039: did not expect clippy to have preferences on the depictions of wyrms

tuf/src/repository/track_repo.rs
```diff
@@ -117,8 +117,7 @@ where
             let mut buf = Vec::new();
             metadata.read_to_end(&mut buf).await?;
 
-            let () = self
```
kellerb: i actually like these unit bindings, makes it very obvious to the reader nothing important is being ignored (i know `must_use` is for that, but sometimes there are exceptions

not that you shouldn't get rid of it here, just be prepared for push back in //src/sys/pkg :)

kellerb (approved):

tuf/src/metadata.rs
erickt: This wasn't *really* a clippy lint, but those `{` always screwed up bracket matching in vim and vscode, so I decided to finally get rid of it.

tuf/src/repository/track_repo.rs
erickt: Oh I generally agree, just in this case the track_repo is a test helper crate, so I wasn't too worried about the return type changing. 

But this seems like a worthwhile lint to disable in general, so I added it.

PR #373, opened 2022-08-25
Delegations have a field named `name`, not `role`

erickt:
For reference, see the [spec].

[spec]: https://theupdateframework.github.io/specification/latest/#delegations

### DIFF
--- a/tuf/src/interchange/cjson/shims.rs
+++ b/tuf/src/interchange/cjson/shims.rs
@@ -368,7 +368,7 @@ pub struct PublicKeyValue {
 
 #[derive(Serialize, Deserialize)]
 pub struct Delegation {
-    role: metadata::MetadataPath,
+    name: metadata::MetadataPath,
     terminating: bool,
     threshold: u32,
     #[serde(rename = "keyids")]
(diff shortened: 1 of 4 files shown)


tuf/src/interchange/cjson/shims.rs
```diff
@@ -2,6 +2,7 @@ use chrono::offset::Utc;
 use chrono::prelude::*;
 use serde_derive::{Deserialize, Serialize};
 use std::collections::{BTreeMap, HashSet};
+use std::convert::{TryFrom, TryInto};
```
JayXon: is `TryInto` needed here? it's included in edition 2021 prelude?

tuf/src/interchange/cjson/shims.rs
```diff
@@ -368,82 +369,109 @@ pub struct PublicKeyValue {
...
+impl From<&metadata::Delegations> for Delegations {
+    fn from(delegations: &metadata::Delegations) -> Delegations {
+        let mut roles = delegations
+            .roles()
+            .iter()
+            .map(Delegation::from)
+            .collect::<Vec<Delegation>>();
+
+        // We want our roles in a consistent order.
+        roles.sort_by(|lhs, rhs| lhs.name.cmp(&rhs.name));
```
JayXon, line 449: nit: use `sort_by_key`?

JayXon (approved):

tuf/src/interchange/cjson/shims.rs
erickt: done

tuf/src/interchange/cjson/shims.rs
erickt: I'd love to be able to use `roles.sort_by_key(|r| &r.name)`, but unfortunately that gives this error:

```
449 |         roles.sort_by_key(|role| &role.name);
    |                            ----- ^^^^^^^^^^ returning this value requires that `'1` must outlive `'2`
    |                            |   |
    |                            |   return type of closure is &'2 MetadataPath
    |                            has type `&'1 shims::Delegation`
```

JayXon: oh I remember this https://github.com/rust-lang/rust/issues/34162

PR #376, opened 2022-08-25
Fix a number of builder issues

erickt:
This fixes a number of bugs in the builder. See each commit message for a description.

### DIFF
--- a/tuf/src/repo_builder.rs
+++ b/tuf/src/repo_builder.rs
@@ -2,16 +2,16 @@
 
 use {
     crate::{
-        crypto::{self, HashAlgorithm, PrivateKey},
+        crypto::{self, HashAlgorithm, PrivateKey, PublicKey},
         database::Database,
         error::{Error, Result},
         interchange::DataInterchange,
         metadata::{
-            Delegations, Metadata, MetadataDescription, MetadataPath, MetadataVersion,
-            RawSignedMetadata, RawSignedMetadataSet, RawSignedMetadataSetBuilder, RootMetadata,
-            RootMetadataBuilder, SignedMetadataBuilder, SnapshotMetadata, SnapshotMetadataBuilder,
-            TargetDescription, TargetPath, TargetsMetadata, TargetsMetadataBuilder,
-            TimestampMetadata, TimestampMetadataBuilder,
+            Delegation, DelegationsBuilder, Metadata, MetadataDescription, MetadataPath,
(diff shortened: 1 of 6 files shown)


tuf/src/repo_builder.rs
```diff
@@ -210,6 +211,56 @@ where
...
+    R: RepositoryStorage<D>,
+{
+    fn root_keys_changed(&self, root: &Verified<RootMetadata>) -> bool {
+        let root_keys_count = root.root_keys().count();
+        if root_keys_count != self.trusted_root_keys.len() {
+            return true;
+        }
+
+        for &key in &self.trusted_root_keys {
+            if !root.root_keys().any(|x| x == key.public()) {
```
JayXon: this is n^2, how many keys are there usually? if it could be large, maybe collect root keys into a HashSet

tuf/src/repo_builder.rs
```diff
@@ -858,22 +858,32 @@ where
     }
 
     fn need_new_targets(&self) -> bool {
-        // If we don't have a database yet, we need to stage the first targets metadata.
+        // We need a new targets if we added and targets.
```
JayXon: I think you mean any targets

tuf/src/repo_builder.rs
```diff
@@ -806,6 +819,18 @@ where
...
+
+            // Insert all the metadata from the trusted snapshot.
+            if self.state.inherit_from_trusted_targets {
+                for (target_path, target_description) in trusted_targets.targets() {
+                    targets_builder = targets_builder
+                        .insert_target_description(target_path.clone(), target_description.clone());
+                }
+
+                if let Some(delegations) = trusted_targets.delegations() {
+                    targets_builder = targets_builder.delegations(delegations.clone());
```
JayXon: not sure if how much difference it makes but since is this a clone, we could avoid doing it here if `self.state.delegations` is some, since it will get overwritten anyway

JayXon (approved):

tuf/src/repo_builder.rs
erickt: Oops, fixed this and some other bad comments too.

tuf/src/repo_builder.rs
erickt: even though the key count should be pretty small, I switched to a hashset to make it not be a problem in extreme cases.

tuf/src/repo_builder.rs
erickt: I ended up fixing this in a later patch.

JayXon (approved):

PR #377, opened 2022-08-30
Tweaks to better interop with go-tuf

lukesteensen:
👋 I've been working on integrating the [Vector](https://github.com/vectordotdev/vector) project with Datadog's TUF/Uptane implementation, and this PR contains the handful of tweaks I've had to make to get everything interoperating happily:

1. A `spec_version` of `"1.0"` is not actually valid according to SemVer. It should probably be `"1.0.0"` everywhere, but I've kept a check for the old value so that it continues to work with existing metadata.
2. The datetime parsing was overly strict relative to the spec, which only specifies ISO8601. Here it's adjusted to use a full RFC3339 parser that handles the slightly differing format that go-tuf can emit.
3. Allowing `*` in paths, which is legal according to the spec and used somewhat heavily when specifying delegations.
4. Adding a couple of `Send` bounds, which isn't related to interop but made the library easier to use.

The `spec_version` change does seem to break the interop tests, but I wanted to make sure we want to move forward with the change before doing the work of regenerating the test data.

If any these seem undesirable or warrant more discussion, I'd be happy to split them into separate PRs.

/cc @zenithar @cedricvanrompay-datadog

### DIFF
--- a/tuf/src/interchange/cjson/shims.rs
+++ b/tuf/src/interchange/cjson/shims.rs
@@ -579,3 +579,51 @@ mod deserialize_reject_duplicates {
         })
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::{parse_datetime, valid_spec_version};
+
+    #[test]
+    fn spec_version_validation() {
+        let valid_spec_versions = ["1.0.0", "1.0"];
+
+        for version in valid_spec_versions {
+            assert!(valid_spec_version(version), "{:?} should be valid", version);
+        }
+
+        let invalid_spec_versions = ["1.0.1", "1.1.0", "2.0.0", "3.0"];
+
+        for version in invalid_spec_versions {
+            assert!(
+                !valid_spec_version(version),
+                "{:?} should be invalid",
+                version
+            );
+        }
+    }
+
+    #[test]
+    fn datetime_formats() {
+        // The TUF spec says datetimes should be in ISO8601 format, specifically
+        // "YYYY-MM-DDTHH:MM:SSZ". Since not all TUF clients adhere strictly to that, we choose to
+        // be more lenient here. The following represent the intersection of valid ISO8601
+        // and RFC3339 datetime formats (source: https://ijmacd.github.io/rfc3339-iso8601/).
+        let valid_formats = [
+            "2022-08-30T19:53:55Z",
+            "2022-08-30T19:53:55.7Z",
+            "2022-08-30T19:53:55.77Z",
+            "2022-08-30T19:53:55.775Z",
+            "2022-08-30T19:53:55+00:00",
+            "2022-08-30T19:53:55.7+00:00",
+            "2022-08-30T14:53:55-05:00",
+            "2022-08-30T14:53:55.7-05:00",
+            "2022-08-30T14:53:55.77-05:00",
+            "2022-08-30T14:53:55.775-05:00",
+        ];
+
+        for format in valid_formats {
+            assert!(parse_datetime(format).is_ok(), "should parse {:?}", format);
+        }
+    }
+}
--- a/tuf/src/interchange/cjson/shims.rs
+++ b/tuf/src/interchange/cjson/shims.rs
@@ -25,7 +25,15 @@ fn parse_datetime(ts: &str) -> Result<DateTime<Utc>> {
 }
 
 fn format_datetime(ts: &DateTime<Utc>) -> String {
-    ts.to_rfc3339()
+    format!(
+        "{:04}-{:02}-{:02}T{:02}:{:02}:{:02}Z",
+        ts.year(),
+        ts.month(),
+        ts.day(),
+        ts.hour(),
+        ts.minute(),
+        ts.second()
+    )
 }
 
 #[derive(Debug, Serialize, Deserialize)]
--- a/tuf/src/interchange/cjson/shims.rs
+++ b/tuf/src/interchange/cjson/shims.rs
@@ -8,7 +8,15 @@ use crate::error::Error;
 use crate::metadata::{self, Metadata};
 use crate::Result;
 
-const SPEC_VERSION: &str = "1.0";
+const SPEC_VERSION: &str = "1.0.0";
+
+// Ensure the given spec version matches our spec version.
+//
+// We also need to handle the literal "1.0" here, despite that fact that it is not a valid version
+// according to the SemVer spec, because it is already baked into some of the old roots.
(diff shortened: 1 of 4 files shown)


tuf/src/interchange/cjson/shims.rs
```diff
@@ -8,23 +8,24 @@ use crate::error::Error;
...
 use crate::Result;
 
-const SPEC_VERSION: &str = "1.0";
+const SPEC_VERSION: &str = "1.0.0";
+
+// Ensure the given spec version matches our spec version.
+//
+// We also need to handle the literal "1.0" here, despite that fact that it is not a valid version
+// according to the SemVer spec, because it is already baked into some of the old roots.
+fn valid_spec_version(other: &str) -> bool {
```
erickt, line 17: Could you add a test to make sure we can still work with metadata that has the old `spec_version: "1.0"`?

tuf/src/interchange/cjson/shims.rs
```diff
@@ -8,23 +8,24 @@ use crate::error::Error;
...
+    other == SPEC_VERSION || other == "1.0"
+}
 
 fn parse_datetime(ts: &str) -> Result<DateTime<Utc>> {
```
erickt, line 21: Thanks, yeah we should use `parse_from_3339(...)` here. I wish we had a formal one to fully parse ISO 8601 since that's what the spec says we could get, but this is at least better than what we currently have.

Could you add a test for this as well, just to make sure we support the old scheme, and what go-tuf can produce?

tuf/src/metadata.rs
```diff
@@ -66,7 +66,6 @@ static PATH_ILLEGAL_STRINGS: &[&str] = &[
     "\"",
     "|",
     "?",
-    "*",
```
erickt, line 69: It'd be helpful if we also had a test here so we don't accidentally break this again.

erickt (changes requested):
Hello! Great patch, thanks! Beyond my comments, it looks like there's some test failures from us changing the `spec_version`. It also might be time to update our golden test files from python-tuf and go-tuf.

--

And sorry that I keep putting out beta versions without cutting a formal release. I think with my delegation fix, and fleshing out the builder more I'm almost ready to commit to not breaking the API. I was tracking that work in https://github.com/theupdateframework/rust-tuf/issues/355. One thing to note though I'm not super confident in our delegation support. We aren't currently using it, so outside of the tests here we don't have a lot of real world use cases. For example, I just found that we were using the wrong field name last week in https://github.com/theupdateframework/rust-tuf/pull/373. I'd love any help trying to harden that.

lukesteensen:
Thanks! I've updated the lib tests, so those should now be passing, but left the interop tests since I'm not sure of the best procedure for generating new fixtures.

As far as testing delegation support, I'll see if some of our TUF experts can have a look and give some feedback. I'm pretty sure we use them at Datadog, but my team is still pretty new to TUF.

trishankatdatadog:
First, nice work!

> As far as testing delegation support, I'll see if some of our TUF experts can have a look and give some feedback. I'm pretty sure we use them at Datadog, but my team is still pretty new to TUF.

Second, yes, we do use delegations for [full verification](https://uptane.github.io/papers/uptane-standard.2.0.0.html#rfc.section.5.4.4.2). I recommend taking a look at how [go-tuf](https://github.com/theupdateframework/go-tuf/blob/314eed46a8b135443ac3a9efe956cf2a6221ee9f/pkg/targets/delegation_test.go) or [python-tuf](https://github.com/theupdateframework/python-tuf/blob/develop/tests/test_updater_delegation_graphs.py) does it.

cedricvanrompay-datadog:
Interesting, it seems that `go-tuf` has it wrong as well for the `spec_version` format: https://github.com/theupdateframework/go-tuf/blob/ac7b5d7bce18cca5a84a28b021bd6372f450b35b/data/types.go#L112

trishankatdatadog:
> Interesting, it seems that `go-tuf` has it wrong as well for the `spec_version` format: https://github.com/theupdateframework/go-tuf/blob/ac7b5d7bce18cca5a84a28b021bd6372f450b35b/data/types.go#L112

If so, would you help us to create a new issue there? Thanks!

trishankatdatadog:
BTW, we don't have [keyid_hash_algorithms](https://github.com/theupdateframework/go-tuf/issues/368) in the metadata here, do we?

lukesteensen:
I've added one more commit here to add a catch-all `additional_fields` property to ~~`TargetsMetadata`~~ all of the top level metadata structs. It turns out that [go-tuf](https://github.com/theupdateframework/go-tuf/blob/ac7b5d7bce18cca5a84a28b021bd6372f450b35b/data/types.go#L214) has a `custom` field here that is not mentioned in the spec. I didn't necessarily want to duplicate that, but the spec does also say that [implementations can use additional fields and they should be maintained](https://github.com/theupdateframework/go-tuf/blob/ac7b5d7bce18cca5a84a28b021bd6372f450b35b/data/types.go#L214), so that's what I did. If someone has a better solution for that I'd be happy to discuss it.

trishankatdatadog (commented):
IDK Rust, but the ideas therein seems good enough to me 👍🏽

erickt:
It looks like we also got some test failures. I'd also like the `additional_fields` to have some tests too. It might be better to factor that out into a separate PR so it doesn't conflate with the other changes.

erickt:
FYI I did find a bug in rust-tuf's delegation https://github.com/theupdateframework/rust-tuf/issues/381.

erickt:
I just realized that switching our metadata generation to building metadata with the spec_version of `1.0.0` would cause old clients to reject the now spec-conformant metadata, so we need to come up with a way to do a graceful migration.

As a stopgap, I factored out some of your fix in #386 to make sure new clients will at least be able to support the proper format once we start producing it.

lukesteensen:
@erickt Thanks so much for taking most of these across the line! Apologies for the radio silence, I had a couple weeks of travel there and some conflicting work priorities. It seems like the `additional_fields` piece, as well as maybe the `Send` bound tweak, are all that's left? This coming week I'm happy to open a more specific PR for just that.

erickt:
Yeah, although I saw and landed a patch that removed a number of unnecessary trait bounds. It might be that inheriting `Send` might not be necessary anymore.

erickt:
I think with merging the additional_fields PR this workstream is complete so I'm going to close this PR. Please let me know if I missed anything though!

PR #379, opened 2022-08-31
Allow expiration duration to be customized

erickt:
Changing the expiration duration is one of the main causes for users to call the `stage_*_with_builder`. This allows more use cases to avoid needing to call those functions by allowing the user to change the expiration duration without forcing a new metadata to be generated.

### DIFF
--- a/tuf/src/repo_builder.rs
+++ b/tuf/src/repo_builder.rs
@@ -474,7 +474,7 @@ where
         self
     }
 
-    /// Sets that the root metadata will expire after after this duration past the current time.
+    /// Sets that the root metadata will expire after this duration past the current time.
     ///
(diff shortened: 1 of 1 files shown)


tuf/src/repo_builder.rs
```diff
@@ -457,6 +474,55 @@ where
         self
     }
 
+    /// Sets that the root metadata will expire after after this duration past the current time.
```
kellerb: two "after"s

kellerb (approved):

tuf/src/repo_builder.rs
erickt: fixed thanks.

PR #382, opened 2022-09-20
FileSystemRepository does not need to be failable

erickt:
This changes the FileSystemRepository already create the metadata and target's parent directories when they are written to disk.

### DIFF
--- a/tuf/src/repository/file_system.rs
+++ b/tuf/src/repository/file_system.rs
@@ -59,26 +59,24 @@ where
     }
 
     /// Build a `FileSystemRepository`.
-    pub fn build(self) -> Result<FileSystemRepository<D>> {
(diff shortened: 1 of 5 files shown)


kellerb (approved):

tuf/src/repository/file_system.rs
```diff
@@ -469,12 +465,11 @@ mod test {
...
                 .metadata_prefix("meta")
                 .targets_prefix("targs")
-                .build()
-                .unwrap();
+                .build();
 
             // test that init worked
-            assert!(temp_dir.path().join("meta").exists());
-            assert!(temp_dir.path().join("targs").exists());
+            //assert!(temp_dir.path().join("meta").exists());
```
kellerb, line 471: delete?

kellerb (approved):

PR #384, opened 2022-09-21
Switch RepositoryStorage::store_* to take &self

erickt:
During our migration to futures, we switched
`RepositoryStorage::store_metadata` and `RepositoryStorage::store_targets` to take `&mut self` to make it easier to reason about changes. This change however prevented us from concurrently changing the repository, which can speed up large commits. This change switches the repositories back to use `&self` to enable this.

One consequence of doing this is we need to change the batch update implementation `commit()` to now be fallible. This is implemented by incrementing a version number on every repository change, and storing the initial version number in the batch. We have a conflict if the versions don't match during commit.

### DIFF
--- a/tuf/src/repository/ephemeral.rs
+++ b/tuf/src/repository/ephemeral.rs
@@ -1,23 +1,40 @@
 //! Repository implementation backed by memory
 
-use futures_io::AsyncRead;
-use futures_util::future::{BoxFuture, FutureExt};
-use futures_util::io::{AsyncReadExt, Cursor};
-use std::collections::HashMap;
-use std::marker::PhantomData;
-
-use crate::error::Error;
-use crate::interchange::DataInterchange;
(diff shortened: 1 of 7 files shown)


tuf/src/repository/ephemeral.rs
```diff
@@ -358,4 +445,85 @@ mod test {
             );
         })
     }
+
+    #[test]
+    fn ephemeral_repo_batch_commit_fails_with_metadata_conflicts() {
```
kellerb, line 450: possibly also test that two concurrent batch updates will cause one to fail

kellerb (approved):

tuf/src/repository/ephemeral.rs
erickt: Good idea, and added some extra impls we can now cover.

Issue #388, opened 2022-09-23
rust-tuf doesn't support delegation glob pathpatterns


erickt:
The TUF spec says delegation [pathpattern](https://theupdateframework.github.io/specification/latest/#pathpattern)s support unix shell-style pattern matching, but that's not supported in rust-tuf yet.

PR #389, opened 2022-09-23
Allow paths to contain `*`

erickt:
From #377, the [TUF spec] allows for paths to contain `*`. According to @lukesteensen, this is routinely used in delegations.

Note though this does not attempt to implement delegation path globs. That will be implemented in #388.

[TUF spec]: https://theupdateframework.github.io/specification/latest/#targetpath

### DIFF
--- a/tuf/src/metadata.rs
+++ b/tuf/src/metadata.rs
@@ -66,7 +66,6 @@ static PATH_ILLEGAL_STRINGS: &[&str] = &[
     "\"",
     "|",
     "?",
-    "*",
     // control characters, all illegal in FAT
     "\u{000}",
     "\u{001}",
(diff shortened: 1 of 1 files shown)


tuf/src/metadata.rs
```diff
@@ -2314,6 +2313,24 @@ mod test {
...
+            "*/some/path/",
+            "some/*/path",
+            "some/*/path/*",
+        ];
+
+        for path in good_paths.iter() {
+            assert!(safe_path(path).is_ok());
+            assert!(TargetPath::new(path.to_string()).is_ok());
+            assert!(MetadataPath::new(path.to_string()).is_ok());
+            assert!(TargetPath::new(path.to_string()).is_ok());
```
kellerb: duplicate

kellerb (approved):

PR #400, opened 2024-02-08
Update *ring* to 0.17 and untrusted to 0.9. 

briansmith:
Intentionally use "0.17" as the version spec instead of the latest version "0.17.7" to help projects depending on rust-tuf who want to go through a two step ring 0.16.20 -> 0.17.0 -> 0.17.7 update process.

Since *ring* 0.17 uses untrusted 0.9 currently, update rust-tuf to use that same version.


briansmith:
CI won't pass until #401 is merged. /cc @erickt

erickt:
A _very_ belated thank you!

PR #402, opened 2024-02-09
Remove the `unstable_rsa` feature.

briansmith:


erickt (approved):

briansmith:
FYI: I updated this PR after it was approved.

briansmith:
Alternatively to this, I could extend #403 to support RSA.

Issue #410, opened 2024-10-15
Integrating tuf-conformance


jku:
We've written a client conformance test suite: https://github.com/theupdateframework/tuf-conformance/

I made a quick-and-dirty attempt at integrating it in this project. There are likely some other issues remaining but these two are currently preventing progress:
* #408 -- the conformance test suite had to choose something as the default key ... and that was ecdsa. Adding a `default-keytype` option to the test suite is a possibility if ecdsa support here is unlikely
* rust-tuf client seems to only accept metadata with spec_version "1.0.0"
  * the specification is mighty unhelpful here with _Adopters are free to determine what is considered a match_
  * but I think only accepting 1.0.0 is not useful when that version is more than 4 years old and  the specification is at 1.0.33 and the spec maintainers attempt to keep changes "forward-compatible". 
  * I _think_ a client accepting 1.0.x is a reasonable choice and something most clients do (unfortunately there is no specification changelog other than git log so I can't promise that there are no incompatibilities)

As an additional note:
* The components in `src/repository/` seem to assume that the metadata and targets come from the same "source": at least for HttpRepository this is incorrect: artifacts (targets) might be downloaded from a different host than metadata files. This is not the case for the conformance test suite currently but it is true for some real world repositories.

I'm not sure when I'll continue with this conformance client project, I thought I'd file this in the interim.

PR #417, opened 2026-06-22
Support `spec_version` `1.0.x` for better interoperability with other TUF implementations

llbxg:
## description

I am evaluating `rust-tuf` as a client-side library for embedded Linux / CIP[^cip] use cases.

As part of that work, I tested a client based on `rust-tuf` with other TUF repositories and server-side implementations.

I found an interoperability issue. The `rust-tuf` client currently accepts `spec_version` values such as `1.0` and `1.0.0`.  
But current TUF metadata in the ecosystem often uses `1.0.x`.

This PR keeps compatibility with the old `1.0` value. It also accepts numeric `1.0.x` values such as `1.0.31`.

I kept this change small.

- `1.0` is still accepted for compatibility with existing metadata
- numeric `1.0.x` is accepted
- unrelated versions such as `1.1.0` and `2.0.0` are still rejected
- this PR does **not** change the match policy to all `1.x.y`

The TUF specification leaves this point open. It says that `spec_version` uses SemVer, but adopters can decide what counts as a match [^tuf].  
Because of that, I am proposing a small interoperability fix here, not a broader policy change.

I also checked a few other TUF projects.

- `tuf-conformance` includes metadata with `spec_version: "1.0.31"` [^conformance]
- `python-tuf` accepts numeric `X.Y` / `X.Y.Z` and matches on major version [^python]
- `tuf-js` seems to use almost the same policy [^js]
- `go-tuf` writes `1.0.31`, and from a quick look at the current source I could not find the same strict `spec_version` check in the main parse path [^go]

There is still a bigger design question: should `rust-tuf` accept `1.x.y` more generally, like `python-tuf` and `tuf-js`?  
I think that question should be decided separately.  
For this PR, I wanted to focus on the immediate interoperability issue reported in #410.

If maintainers prefer a different direction, I am happy to adjust this PR or split follow-up interoperability changes into separate PRs.

## verification

- updated the existing `spec_version` validation test
- added a regression test that root metadata deserialization accepts `spec_version: "1.0.31"`
- ran targeted tests for `spec_version`
- ran the full test suite

<details><summary>Click to expand</summary>

```sh
cargo test -p tuf --lib pouf::pouf1::shims::test::spec_version_validation -- --exact
```

```plaintext
running 1 test
test pouf::pouf1::shims::test::spec_version_validation ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 166 filtered out; finished in 0.00s
```

```sh
cargo test -p tuf --lib metadata::test::deserialize_json_root_accepts_1_0_x_spec_version -- --exact
```

```plaintext
running 1 test
test metadata::test::deserialize_json_root_accepts_1_0_x_spec_version ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 166 filtered out; finished in 0.00s
```

```sh
cargo test -q
```

-> all tests passed

</details>

## REF

- #410
- #386
- !377

[^cip]: https://cip-project.org/
[^tuf]: https://github.com/theupdateframework/specification/blob/v1.0.31/tuf-spec.md?plain=1#L690
[^conformance]: https://github.com/theupdateframework/tuf-conformance/blob/eeb82bff9ce7e19133d24bc7bd44a7e8908546ed/tuf_conformance/static_data/tuf-on-ci-0.11/metadata/1.root.json#L60
[^python]: https://github.com/theupdateframework/python-tuf/blob/9a3c3046d6ffdc9d90ec21ce5237721bcd985652/tuf/api/_payload.py#L137
[^js]: https://github.com/theupdateframework/tuf-js/blob/a399ce87d8d908f64d53696a673420193424e0de/packages/models/src/base.ts#L41
[^go]: `metadata/types.go`, `metadata/metadata.go`, `metadata/marshal.go`


tuf/src/pouf/pouf1/shims.rs
```diff
@@ -14,12 +14,35 @@ use {
...
+    identifier == "0" || !identifier.starts_with('0')
+}
+
+// Ensure the given spec version stays within the supported TUF 1.0.x line.
 //
 // We also need to handle the literal "1.0" here, despite that fact that it is not a valid version
 // according to the SemVer spec, because it is already baked into some of the old roots.
 fn valid_spec_version(other: &str) -> bool {
-    matches!(other, "1.0" | "1.0.0")
+    if other == SPEC_VERSION {
```
erickt: Thanks for this! Looking through your research, I think we should follow the spec and just parse the string as a semver string, preferably using the `semver` crate. With something like:

```
const SPEC_VERSION = semver::Version::new(1, 0, 0);
...
fn valid_spec_version(version_string: &str) -> bool {
    /////////////////////////////////////////
    // TUF-1.0.34 §4.3 (https://theupdateframework.github.io/specification/v1.0.34/#spec_version):
    //
    //     A string that contains the version number of the TUF specification. Its format follows
    //     the [Semantic Versioning 2.0.0 (semver)](https://semver.org/spec/v2.0.0.html)
    //     specification. Metadata is written according to version "spec_version" of the
    //     specification, and clients MUST verify that "spec_version" matches the expected version
    //     number. Adopters are free to determine what is considered a match (e.g., the version
    //     number exactly, or perhaps only the major version number
    //     (major.minor.fix).https://theupdateframework.github.io/specification/v1.0.34/#file-formats-root

    let version = semver::Version::parse(version_string) else {
        // Support parsing `1.0` for legacy versions
        return version_string == "1.0";
    };

    version.major == SPEC_VERSION.major
}
```

At this point I think we can switch our spec version to 1.0.0, and eventually update the metadata version to 1.0.x once I finish updating support for the latest spec.

erickt (changes requested):

tuf/src/pouf/pouf1/shims.rs
llbxg: Thanks for the suggestion. 

I updated the implementation to use the `semver`
crate and defined the supported spec version as `Version::new(1, 0, 0)`.

As part of this change, metadata generated by the test helpers now uses
`"1.0.0"` instead of `"1.0"`. I updated the affected test fixtures and
expected signatures accordingly. I also changed the spec version to `"1.0.0"`
in tests that are intended to fail for another validation reason, so that they
continue to exercise the intended failure path.

To preserve compatibility with legacy metadata, I added
`deserialize_json_root_accepts_legacy_1_0_spec_version`, which verifies that
we still accept `"1.0"`.

The interop metadata fixtures also needed to be regenerated because the
updated `spec_version` changes the signed metadata. I regenerated them using
`interop-tests/tests/metadata/regenerate-metadata.sh`. The interop tests pass
for both consistent snapshot configurations.

Please let me know if you see any issues with these changes. I would be happy
to address them.

erickt (approved):

erickt:
Could you do a `cargo fmt` for this? Otherwise LGTM.

llbxg:
My apologies, I forgot to run `cargo fmt` before submitting the changes.  
I see that you've already addressed it. Thank you very much for your support.

PR #419, opened 2026-07-14
Extract HttpRepository into tuf-hyper, port to hyper 1.x

erickt:
This extracts HttpRepository into its own crate, so its easier for users to not pull in the http stack unless it is necessary. It also updates hyper to 1.x, and adds some examples on how to create a client using tuf-hyper.


tuf-hyper/examples/client-smol.rs
```diff
@@ -0,0 +1,275 @@
...
+    inner: smol::net::TcpStream,
+}
+
+impl Read for SmolStream {
+    fn poll_read(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        mut cursor: ReadBufCursor<'_>,
+    ) -> Poll<io::Result<()>> {
+        let buf = unsafe { cursor.as_mut() };
```
kellerb: add a comment for this unsafe like "No bytes of `buf` are un-initialized by this function"

kellerb (approved):

tuf-hyper/examples/client-smol.rs
erickt: Done! And found there's a better way coming in the next hyper release too.

